PR #27, opened 2026-02-26
tests: Test correctness

XbzOnGit:
Add E2E correctness tests comparing with original HF transformers


Samfisheryu:
A few changes bundled in:

1. **Dead `MON_NATIVE_*` monkeypatches** — PR #24 already removed these env vars, the `monkeypatch.setenv` calls were silently no-ops. Switched to `MonitoringConfig(advance=AdvanceConfig())`.

2. **Wrong import path for `HostEngineConfig`** — was imported from `monitoring._native_engine` (doesn't exist there), fixed to import from `monitoring`.

3. **`get_num_layers_from_config` inside the per-request loop** — moved before the loop, added `assert n_layers == num_layers` to cross-check against model config.

4. **Triple-quoted string as a comment** — was used to disable ~20 lines of unfinished `hook_final_ln` code. Replaced with a single `# TODO`; needs a follow-up issue.

5. **Split into `tests/correctness/` subpackage** — 1266-line monolith broken into `db_reader`, `tensor_utils`, `hf_reference`; main test file down to ~350 lines.

6. **Per-assertion subtests** — added `pytest-subtests`, each comparison wrapped in `with subtests.test(msg=...)`. With default config (batch_size=4, GPT2) this gives ~164 individually reported PASS/FAIL items instead of one big traceback.

Verified: `E2E_HF_DROP_LAST_TOKEN=1 pytest -v -s tests/test_e2e_correctness_vs_hf.py` passes.

Samfisheryu (approved):
LGTM

Samfisheryu:
Add an issue of final logits

PR #30, opened 2026-03-03
Merge/performant monitor

zhaoyboo:


Samfisheryu:
QWEN3 supported? Give a benchmark on Qwen3-8B performance

Samfisheryu:
I test the dual graph path and found 2 separate issues.

  Using `e763f9f4f` as the repro / test case:

1. In the current production path, phase 1 records addresses, then `finalize_dual_frame()` builds aliases, then
`disable_record()` re-traces phase 2 graphs.
That means replay is running on a different graph variant from the one that produced the alias addresses.

In a real prefill + decode setup, this breaks for `hook_embed`:
- step 0 is correct
- after that it becomes stale
- in our E2E repro it shows up as “all rows equal row 0”

This is not a D2H bug. The alias itself is already stale before D2H.

2. If I keep `record` on and skip phase 2 retrace, `hook_embed` becomes correct again.
But then the compiled graph is different enough that numerics can change:
- for some prompts, token ids diverge
- for others, token ids match but hidden states are no longer bitwise equal

So right now there are really 2 modes:
- `record off` production path: semantic alias bug
- `record on` path: no stale alias, but different compiled graph / numeric drift

The existing integration test proves:
- D2H == alias read (bit-exact)
but didn't prove correctness

PR #40, opened 2026-04-10
[Core] Replace native backend with ring transport + full TP support

### LINKED ISSUE #34: Consolidate ring buffer implementation and clean dead code paths
Part of #32

## Problem

Ring transport was added on top of the existing native backend. Over time, ring-related branches were added into `HookedRootModule` methods that ring never actually calls in production.

## Dead code in hook_points.py

Ring's actual call chain (both HF and vLLM paths) only touches:
- `HookPoint.forward()` → `torch.ops.ring.producer()`
- `HookPoint.name` setter (converts hook name → ring type/id)
- `HookPoint.enabled` flag

It never enters these `HookedRootModule` methods:
- `prepare_monitoring()` — ring enters and immediately returns
- `run_with_cache()` — ring branches at lines 717-728 and 780-782 are unreachable; HF ring path uses `generate_with_monitoring`, vLLM uses `DMXGPUWorker.execute_model`
- `get_caching_hooks()` — ring skip logic at lines 1004-1006 is unreachable for the same reason

All `_using_ring_transport` / `_using_native_backend` if/else branches in these methods are dead code for ring mode.

## Hook type constants triplicated

`HOOK_TYPE_RESID_PRE = 0`, `HOOK_TYPE_LN1 = 1`, etc. are defined in three places:
- Python: `ring_transport.py` (lines 51-71)
- C++ enum: `csrc/ring/tensor_meta.h` (lines 28-50)
- C++ display names: `csrc/ring/tensor_meta.h` (lines 56-67)

Plus a legacy name mapping dict in `ring_transport.py` (lines 221-242). Adding a new hook type requires synchronized edits to all four locations. Should define once and share.

## No-op forward hooks

`_make_ring_hook()` in `ring_transport.py` creates empty hooks that do nothing. `install_ring_hooks()` registers them but the actual work is in `HookPoint.forward()`. Dead code.

## Tasks

- [x] Remove dead ring branches from `HookedRootModule` (`run_with_cache`, `prepare_monitoring`, `get_caching_hooks`)
- [x] Single source of truth for hook type constants
- [x] Remove no-op forward hook registration (`_make_ring_hook`, `install_ring_hooks`)

XbzOnGit:
## Purpose

Replace the monitoring data path with a GPU ring buffer pipeline. The old native backend (pin-pool cudaMemcpy, Python-side slicing, GIL-coupled DB writes) is removed entirely. The new ring transport does D2D copy on GPU, drain + slice on CPU, and ClickHouse insert — all in C++ without the GIL. Also adds tensor parallelism for both vLLM and HF paths, and consolidates all hook definitions into a single C++ table.

### Ring transport pipeline

- **GPU**: `ring::producer` CUDA kernel — coalesced D2D copy into GPU payload ring. Fires as `torch.ops.ring.producer` (void + effectful custom op, survives torch.compile DCE, works under CUDA graphs, currently requires `VLLM_DISABLE_COMPILE_CACHE=1`).
- **CPU drain thread**: polls task ring entries (managed memory, no PCIe), copies payload via cudaMemcpyAsync to pinned staging.
- **p2p thread**: slices per-request token ranges using TensorMetaFifo metadata, formats ClickHouseRows, submits to DMXHostEngine insert queue.
- **Ring primitives**: payload_ring (two-span wrap), task_ring (producer __threadfence + volatile publish, consumer spin + release), task_entry (64B cache-line aligned, sentinel-based), ring_alloc (RAII device alloc), pinned_staging (monotonic ring buffer).

### Tensor parallelism

- **vLLM** (multi-process): `filter_by_tp_rank()` (rank 0 keeps all, rank 1+ sharded only). Users must call `collective_rpc("stop_monitoring")` before terminating vLLM to flush ring data within vLLM's shutdown deadline.
- **HF** (torchrun + NCCL): TP detected via `torch.distributed`, per-rank ClickHouse tables, per-rank ring engines.
- **Shape computation**: `_compute_hook_shape()` divides by tp_size for sharded hooks, handles GQA kv_heads.
- **null_mode race fix**: `cudaDeviceSynchronize` before/after `cudaMemcpyToSymbol` to prevent stream race during vLLM warmup.

### Unified HOOK_DEFS (single source of truth)

All hook properties in one C++ table (`tensor_meta.h`), exposed via pybind11. Python auto-derives all sets at import time. No hardcoded hook-type sets remain. Adding a new hook = one enum value + one table row.

| Column | Type | Purpose |
|--------|------|---------|
| `id` | `HookType` enum | Compile-time constant (0-20) |
| `act_name` | `const char*` | ClickHouse column name |
| `short_name` | `const char*` | Python selection key |
| `per_layer` | `bool` | `blocks.<L>.<act>` vs global |
| `group` | `HookGroup` int | `ATTN` / `MLP` / `OTHER` |
| `tp_sharded` | `bool` | Pre-all-reduce tensor |
| `shape_class` | `ShapeClass` int | Shape formula selector |
| `pp_stage` | `PpStage` int | `ANY` / `FIRST` / `LAST` |

### Code cleanup

Deleted the entire legacy native backend: `native_engine.cpp/h`, `engine_core.cpp`, `api_submit.cpp`, `hooks.cpp`, `slice.cpp`, `future_process.cpp/h`, `unified.cpp`, `engine_utils.cpp/h`, `nvtx_shim.h`, `dmx_host_utils.cpp`, `task.py`, `utils.py`, `NativeMonitoringEngine`, `BackendFuture`, `NativePartialSealConfig`, `AdvanceConfig`, dead engine params, dead stats, 26 legacy benchmark/test files.

## Test Plan

**Transport correctness only.** The compare model has both `ring::producer` (→ ClickHouse) and `.copy_()` (→ pre-allocated buffers) in the same compiled graph. After `generate()`, per-step buffer slices saved to disk, then bitwise-compared against ClickHouse rows. This verifies that the ring transport pipeline (D2D copy → drain → p2p slice → ClickHouse insert) delivers identical bytes to what the model actually computed. It does **not** test whether the hooks alter model numerics — that is a separate concern (hook numeric impact evaluation) not covered by this PR.

```bash
# HF transport correctness (TP=1)
CUDA_VISIBLE_DEVICES=0 bash tests/run_tp_compare_hf.sh gpt2 eager 1
CUDA_VISIBLE_DEVICES=0 bash tests/run_tp_compare_hf.sh qwen3 cudagraph 1

# HF transport correctness (TP=2)
E2E_GPUS=0,1 bash tests/run_tp_compare_hf.sh qwen3 eager 2

# vLLM transport correctness
bash tests/run_tp_compare_vllm.sh qwen3 eager 1
bash tests/run_tp_compare_vllm.sh qwen3 cudagraph 2

# Full regression (15 tests)
bash tests/run_regression.sh

# Ring C++ unit tests
cd tests/ring && make && ./build/test_rings && ./build/test_producer && ./build/test_ring_engine
```

## Test Result

All bitwise exact (0 failures across all configurations):

| Path | Model | Mode | TP | Subtests |
|------|-------|------|----|----------|
| HF | GPT-2 | eager | 1 | 5,568 |
| HF | GPT-2 | CUDA graph | 1 | 5,568 |
| HF | Qwen3-0.6B | eager | 1 | 12,704 |
| HF | Qwen3-0.6B | CUDA graph | 1 | 12,704 |
| HF | Qwen3-0.6B | eager | 2 | 18,976 |
| HF | Qwen3-0.6B | CUDA graph | 2 | 18,976 |

Per-layer hooks (14): resid_pre, ln1, q, k, v, attn_scores, pattern, z, attn_out, resid_mid, ln2, mlp_in, mlp_post, mlp_out.
Global hooks: token_ids, embed, [pos_embed], resid_final, final_ln, final_logits.
TP-sharded (7/layer): q, k, v, z, attn_scores, pattern, mlp_post.
All subtests = 8 steps × 4 batch × hooks_per_step.

Ring C++ unit tests: 221 + 72 + 6 = 299 pass.


XbzOnGit:
Resolves #34

PR #42, opened 2026-05-03
[Refactor] BackendAdaptor: unify HF + vLLM under one contract; HF post-EOS strip

### LINKED ISSUE #33: Decouple HF and vLLM framework logic from core monitoring engine
Part of #32

## Problem

HF and vLLM specific logic is embedded in core engine files. Capacity check, step bytes computation, and step prep sequence are duplicated between `generate.py` (HF) and `vllm_integration.py` (vLLM). Adding a new framework requires modifying core code.

### Current coupling
- `generate.py` (1168 lines) — entirely HF-coupled (StaticCache, CompileConfig, past_key_values, attention mask decoding)
- `vllm_integration.py` (434 lines) — entirely vLLM-coupled (scheduler_output parsing, CUDA graph padding, architecture remap, force_eager monkey-patch)
- `engine.py` `_prepare_ring_step` — hardcoded to HF kwargs (input_ids, attention_mask, past_key_values)
- `ring_transport.py` — vLLM-specific preset (`vllm-full`) and `batch=0` flattened-batch handling mixed in

### What to do
- Extract shared core logic (capacity check, step bytes calc, step prep sequence) into a unified interface in `monitoring/engine.py`
- Move HF-specific wrapper to `integration/hf_adapter.py`
- Move vLLM-specific wrapper to `integration/vllm_adapter.py`
- Adapters translate framework-specific inputs and call the unified core interface

XbzOnGit:
## Purpose

Unify the HF and vLLM monitoring paths under one `BackendAdaptor` abstract base. Pre-refactor, HF orchestration lived in
`monitoring/engine.py` + `monitoring/generate.py`; vLLM in `monitoring/vllm_integration.py`. This PR factors the common driver into `monitoring/adaptor_base.py` + lands two thin concrete subclasses under `integration/`. Adds a working HF post-EOS strip, removes framework-name knowledge from `monitoring/ring_transport.py`, deletes both shims.

### `BackendAdaptor` + concrete adapters

`monitoring/adaptor_base.py` defines the abstract surface (`detect_model_shape`, `detect_parallel_ranks`, `is_pp_first/last`, `build_step_context`, `on_capacity_exceeded`, optional `adapt_for_cpu_direct`) and shared `attach_model` + `before_forward` driver. Per-step flow defined once: `build_step_context -> _step_bytes -> prepare_step [-> adapt_for_cpu_direct + on_capacity_exceeded if result==2] -> set_step_context -> pre_push_all_metas`.

- `HFAdaptor` (`integration/hf_adapter.py`): per-call batch state (request IDs, starts, finished latch, prefill kv_offsets),
  `_prepare_wrapper` install on `prepare_inputs_for_generation`, prefill left-pad strip, post-EOS strip.
- `VLLMAdaptor` (`integration/vllm_adapter.py`): scheduler-output -> packed `StepContext`, `cudagraph_dispatcher._bs_to_padded_graph_size` lookup, `force_eager_next_batch` flag wrapping `_determine_batch_execution_and_padding`, vLLM TP world-size resolution.
- `integration/model_shape.py` hosts `_make_model_shape_from_hf_config`shared by both.

### HF post-EOS strip

HF batched `generate()` is lockstep -- finished requests keep producing forward activations (HF coerces the next token to
`pad_token_id` once a request hits EOS) until all batch members stop. Pre-PR those rows polluted ClickHouse. `HFAdaptor` keeps a per-request `_batch_finished` latch and detects EOS via one `input_ids[:, -1].tolist()` per decode step (single GPU->CPU sync, B bytes; matches HF's own per-step stop-criteria sync).
Detection runs one step late by construction so the EOS-producing activation is captured; subsequent steps emit zero-length `(start_i, start_i)` ranges. Sync skipped when `_eos_token_ids` is empty or `no_strip_right_pad=True`. EOS resolution chain on `attach_model`: explicit kwarg -> constructor kwarg -> `model.generation_config.eos_token_id` -> `model.config.eos_token_id` -> empty `frozenset()`. Multi-EOS via `frozenset[int]`. vLLM does NOT need this -- its scheduler removes finished requests from `scheduler_output.num_scheduled_tokens` before the next step.

### Strip flags + engine cleanup + renames

Flags on `HFAdaptor.__init__` / `attach_model` (per-call override when explicitly passed) / both HF entry points: `no_strip_left_pad=False` (prefill), `no_strip_right_pad=False` (post-EOS), `eos_token_id=None`.

`MonitoringEngine` final surface: `__init__`, `enable_ring_transport` (now returns `RingTransport`), `next_auto_group_id`, `close`. Deleted `_sync_hook_debug_flag` / `_debug_enabled` / `_no_strip` propagation; sharp-deleted orphaned `MonitoringConfig.debug` / `MonitoringConfig.no_strip` fields. NVTX via monitoring.hook_points.set_monitoring_debug(True)` directly.

`generate_greedy` -> `generate_greedy_with_monitoring`. `monitoring/generate.py` and `monitoring/vllm_integration.py` shims deleted; ~15 import sites under `tests/` / `benchmark/` / `example/` migrated; 5 worker_cls strings updated.


Samfisheryu:
Ring transport default / silent fallback fix

Problem: after this refactor there is no non-ring monitoring backend left, but MonitoringEngine could still be constructed with _ring_transport=None. In the HF path, generate_with_monitoring() would then skip HFAdaptor setup and silently run bare model.generate(), which is worse than failing because users think monitoring is active.

Fix: c34eac31e makes ring transport enabled by default from MonitoringEngine.__init__ using a default RingConfig, still allows custom ring_config injection, and makes enable_ring_transport() replace an existing ring cleanly. generate_with_monitoring() now raises if model.monitoring_engine is missing or ring transport has been explicitly disabled, so it cannot silently run unmonitored.

Samfisheryu:
vLLM step logging fix

Problem: VLLMAdaptor.build_step_context printed one line per step/request on the normal production path.

Fix: 84d569eda gates that existing debug print behind the existing RING_DEBUG_STEP flag, matching the HF adapter's step-debug behavior.

Samfisheryu:
Followed up on the debug-print review in 76103ba7c: both HFAdaptor and VLLMAdaptor now read RING_DEBUG_STEP once during adaptor initialization and cache it on the instance. The per-step hot path now checks the cached flag instead of calling os.environ.get(...) for every step/request.

Samfisheryu:
### Benchmark cleanup and smoke validation

Pushed in `4e62a5ba8` (`fix: refresh benchmark monitoring entrypoints`).

#### What changed

- Removed stale benchmark-only scripts that were no longer runnable after the adaptor refactor:
  - `benchmark/scripts/debug_monitoring.py`
  - `benchmark/test_partition.py`
- Updated `benchmark/scripts/hf_monitoring_generate.py` for the new monitoring API:
  - removed the deleted `MonitoringConfig(debug=...)` kwarg
  - moved `BENCH_NVTX` to `set_monitoring_debug(...)`
  - wired `--ring-payload-mb`, `--ring-pinned-mb`, and `--ring-task-entries` into `MonitoringEngine`
  - removed the old `engine.clear_completed_results()` cleanup path
  - fixed direct script execution by adding the repo root to `sys.path`, since `run_bench.py` invokes this script by file path
- Fixed `benchmark/run_bench.py --no-db` so it now runs only HF + `monitoring_nodb`. It no longer runs no-db twice or stores a no-db result under the DB `monitoring` field.

#### Validation

Ran py-compile over the remaining benchmark Python entrypoints.

Smoke test on GPU0:

```text
model: gpt2
prompts: benchmark/data/prompts_top10.txt
batch_size: 1
max_new_tokens: 1
mode: --no-db
ring payload/pinned: 512 MB
```

Direct `hf_monitoring_generate.py` completed and wrote JSON:

```text
monitoring no-db: 0.5645s, 17.71 tok/s
```

`python -m benchmark.run_bench --no-db` completed with:

```text
HF:                0.3921s, 25.50 tok/s
monitoring_nodb:  0.5254s, 19.03 tok/s
monitoring:       null
```

The smoke run is intentionally tiny, so the timing is only a functional sanity check, not a real performance number.

Samfisheryu (approved):
Approved. The adaptor refactor is aligned with the issue direction, and the follow-up fixes addressed the runtime blockers I found: ring transport is enabled by default with explicit failure instead of silent no-monitoring, vLLM step logging is gated and cached, and the stale benchmark monitoring entrypoints have been refreshed and smoke-tested.

PR #52, opened 2026-06-01
Log DMI ClickHouse sink failures

Samfisheryu:
Summary:
- Log ClickHouse insert stage initialization failures with remediation guidance and connection details.
- Log the first P2P submit failure instead of silently suppressing host-engine submit exceptions.

Validation:
- git diff --check
- conda run --no-capture-output -n agent-dmi make -C HF_Prometheus/monitoring
- Manual missing-database smoke test confirmed the new ClickHouse init error log.


XbzOnGit (approved):
LGTM

PR #56, opened 2026-06-11
(#1) test: marker taxonomy + skip guards + categorize

SieDeta:
## Summary

This PR introduces a structured test taxonomy, requirement guards, and test categorization improvements to make the test suite more maintainable and CI-friendly.

### 1. Test Marker Taxonomy

Added a standardized set of pytest markers in `pyproject.toml`:

* `cpu`
* `gpu`
* `multi_gpu`
* `e2e`
* `clickhouse`
* `vllm`
* `hf`
* `ring_native`
* `slow`
* `manual`
* `numeric`

Default test collection behavior:

* Excludes `manual` tests.
* Ignores the following directories via `norecursedirs`:

  * `tests/tools`
  * `tests/ring`
  * `.venv`
  * `integration`
  * `build`

---

### 2. Requirement-Based Skip Guards

Added `tests/_requirements.py` containing CPU-importable helper functions that fail gracefully with clear skip reasons instead of crashing during import.

Available guards:

* `require_cuda()`
* `require_gpus()`
* `require_clickhouse()`
* `require_vllm()`
* `require_model_cache()`
* `require_nvcc()`

---

### 3. Test Categorization

Categorized the existing test suite according to execution requirements:

* **Pure CPU contract/unit tests**

  * Marked with `cpu`

* **vLLM/Hugging Face E2E suites**

  * Explicitly marked with combinations of:

    * `gpu`
    * `vllm`
    * `clickhouse`
    * `e2e`
    * `hf`

* **`test_gpt2_parity`**

  * Uses real GPT-2 weights via `from_pretrained`
  * Marked with:

    * `hf`
  * Protected by:

    * `require_model_cache("gpt2")`
  * No longer classified as a CPU-only test

---

#### `test_producer_chunked_schema.py`

Split into:

* **CPU tests**

  * Operation registration and validation logic

* **GPU tests**

  * CUDA device smoke tests

#### `test_per_hook_isolation.py`

Separated into:

* **CPU tests**

  * Source-patch unit test classes

* **GPU tests**

  * Existing slow GPU sweep tests
  * Retains the `slow` marker

---

### 5. Manual Test Utilities Reorganization

Moved manual shell wrappers into:

```text
tests/tools/
```

Additional changes:

* Added accompanying documentation (`README`).
* Fixed repository root path calculations:

```text
dirname/..       → dirname/../..
```

* Updated sibling script references to account for the new directory depth.

### Default CPU Suite (recommended for PR validation and local development)

```bash
python -m pytest -m "not gpu and not e2e and not manual" -q
```

## Verification

* ✅ All markers are registered without unknown-marker warnings.
* ✅ CPU-only selection correctly deselects 32 GPU/E2E tests.
* ✅ Changes to import-sensitive test files are purely additive (marker annotations only).

### Unbuilt Transformers Submodule

The vendored `transformers` submodule has not been built, causing:

```text
transformers.models.qwen2_moe_compare
```

to be unavailable.

```
`tests/test_moe_v1_routing_hooks.py` imports `transformers.models.qwen2_moe_compare`
and `transformers.models.qwen2_moe_p`, but these packages do not exist in any commit
or branch of the vendored `transformers-DMI` fork. The pinned submodule commit
`ba80bb2570` only has upstream `qwen2_moe`, so the test fails at collection time.
The model packages were never pushed to the fork and must be committed before the
submodule pin can be updated.
```


Samfisheryu:
Looks like you may have accidentally backward the submodule pointers.

Current `main` points to:
  - `integration/transformers`: `ce5095a`
  - `integration/vllm`: `f31edd8`

  This PR points them to older commits:
  - `integration/transformers`: `ba80bb2`
  - `integration/vllm`: `a8fc477`

Could you please pull/rebase on latest `main` and restore these submodule pointers? If this PR does not intend to change submodules, ideally they should disappear from the diff.

integration/transformers
Samfisheryu, line 1: please pull to the newest main and submodule

pyproject.toml
```diff
@@ -32,6 +32,21 @@ monitoring = [
 include = ["monitoring*", "benchmark*", "integration"]
 
 [tool.pytest.ini_options]
+# `manual` tools and native-ring sources are not collected by default. The
+# default suite is the CPU contract/unit tests:
+#   python -m pytest -m "not gpu and not e2e and not manual" -q
+addopts = "-ra -m 'not manual'"
```
Samfisheryu, line 38: comment says default should be cpu only, but here is "not manual"?
maybe `addopts = "-ra -m 'not gpu and not e2e and not manual'"`?

SieDeta:
Ah yes, thanks for mentioning it. I'll fix the other PR too.

tests/_requirements.py
```diff
@@ -0,0 +1,128 @@
...
+from __future__ import annotations
+
+import importlib.util
+import os
+import shutil
+import socket
+
+import pytest
+
+__all__ = [
```
Samfisheryu, line 32: did you use these functions anywhere?

tests/test_tp_shapes.py
```diff
@@ -6,6 +6,8 @@
 import pytest
 import torch
 
+pytestmark = pytest.mark.cpu
+
 from monitoring.ring_transport import (
```
Samfisheryu: this test will import ring_transport which requires building .so. So it's not a cpu only test

Samfisheryu (commented):
be aware some tests marked as `cpu` still require the native `.so` to be built, e.g. imports through `monitoring.ring_transport` / `_load_extension()`. Might be worth double-checking .

integration/transformers
Samfisheryu: also other PRs as well

SieDeta: Yep already done it with other PRs too

pyproject.toml
SieDeta: Sorry, there's been misleading between the comment and the config. Basically a bare pytest runs whatever your machine can support and self-skips the rest.

- On a CPU-only box (no CUDA available) -> CPU-only. The GPU/e2e/vllm tests get collected but each one hits its guard (if not cuda: pytest.skip()) and skips itself. So in practice, yes, you get a CPU-only run, but they just show up as skipped instead of being hidden.
- On a GPU box (NVIDIA GPU + CUDA available) -> it runs the CPU tests plus the GPU/e2e/vllm tests, because those guards now pass. So it is not CPU-only there 

Should I to add an option before the test (config) and user can select which test they want to run (cpu/gpu/etc)? If user not selecting, the default will be cpu.

Example:

```
  pytest                  # default -> cpu suite
  pytest --suite gpu      # gpu tests
  pytest --suite e2e      # end-to-end
  pytest --suite all      # everything the machine supports
```

tests/_requirements.py
SieDeta: On the current PR, only require_model_cache is used (in test_gpt2_parity.py). 
The rest is being used on PR #59 and PR #60.

tests/test_producer_chunked_schema.py
```diff
@@ -21,18 +21,35 @@ def setup_module(module):  # noqa: D401 -- pytest hook
     _load_extension()  # ensure .so loaded -> three ring ops registered
```
XbzOnGit: This file still is not CPU-only even though several tests are marked `cpu`: `setup_module()` unconditionally calls `_load_extension()`, so a checkout without the native extension built fails before the `cpu` tests can run.

pyproject.toml
XbzOnGit: The documented/default CPU selection should also account for `hf` tests. `-m "not gpu and not e2e and not manual"` still selects `tests/test_gpt2_parity.py`, because it is marked `hf` and only skipped when `gpt2` is not cached.

tests/test_producer_chunked_schema.py
Samfisheryu: Fixed in bd5088c5e. `test_producer_chunked_schema.py` is no longer marked `cpu`; it is marked `native_backend`, and `_load_extension()` now skips cleanly if the native `.so` is unavailable.

pyproject.toml
Samfisheryu: Fixed in bd5088c5e. The documented CPU gate is now `python -m pytest -m "cpu" -q`; `test_gpt2_parity.py` is marked `hf` + `framework_fork`, so it is not selected by the CPU-only gate.

tests/test_tp_shapes.py
Samfisheryu: fixed

Samfisheryu:
Note for future test categorization after bd5088c5e: I added two non-CPU categories so we do not accidentally pull heavy dependencies into the CPU-only gate.

- `native_backend`: requires the DMI native backend `.so` to be built/importable. This may still run on CPU, but it is not CPU-only from a CI/bootstrap perspective. Example: tests that call `_load_extension()` or require registered `torch.ops.ring.*` ops.
- `framework_fork`: requires vendored/modified framework forks to be importable, such as modified `transformers` modules or vendored vLLM integration modules. These should also stay out of the CPU-only gate.

The lightweight CPU gate should remain explicit: `python -m pytest -m "cpu" -q`. If later PRs recategorize tests, please keep anything requiring either native `.so` or framework forks out of `cpu` and mark it with the more specific category instead.

Samfisheryu:
Follow-up in 2c133decc: moved the remaining ring_transport/import-chain tests out of the `cpu` gate and into `native_backend`:

- `tests/test_adapter_protocol.py`
- `tests/test_hf_eos_strip.py`
- `tests/test_hook_spec_flags.py`
- `tests/test_tp_shapes.py`

These tests do not require GPU, but they currently require the native hook-definition layer from `monitoring.ring_transport`, so they are not part of the no-native-build CPU gate. I also made the imports collection-safe so `pytest -m "cpu"` on a checkout without the native `.so` does not fail during collection.

Verified locally:
- `python -m pytest -m cpu -q` -> `38 passed, 98 deselected`
- the four `native_backend` files above -> `54 passed` on a local env with the `.so` available

PR #57, opened 2026-06-11
(#2) test: normalize env vars + harden source-patch

SieDeta:
## Summary

This PR standardizes E2E hook configuration, strengthens source-patching isolation guarantees, and improves test robustness without introducing regressions.

### Part A — Environment Variable Normalization

Standardized **`E2E_HOOK_SELECTION`** as the single public configuration interface for E2E tests.

Internally, all subprocesses now translate this value to the runtime contract expected by the monitoring layer:

```text
E2E_HOOK_SELECTION → DMX_HOOK_SELECTION
```

All usages of the legacy `E2E_HOOKS` variable have been removed.

#### Updated Components

* **`test_vllm_identical.py`**

  * Now reads `E2E_HOOK_SELECTION` (previously `E2E_HOOKS`)
  * Removed duplicated docstring content

* **`test_vllm_rowcnt.py`**

  * Now correctly honors the documented `E2E_HOOK_SELECTION` behavior

  * Constructs a dedicated `sub_env` that injects:

    ```text
    DMX_HOOK_SELECTION=<selected hooks>
    ```

    for both subprocesses

  * Previously, the test documented this behavior but passed `os.environ` directly without performing the translation

* **Shell Utilities**

  * Updated all scripts to use `E2E_HOOK_SELECTION`:

    * `tools/verify_vllm.sh`
    * `tools/identical_vllm.sh`
    * `tools/run_qwen2_moe_vllm_pipeline.sh`

---

### Part B — Source-Patch Hardening

Introduced a stronger isolation contract for tests that patch vendored comparison modules.

#### New: `isolated_hook(...)`

Added:

```python
isolated_hook(framework, model_key, hook)
```

in:

```text
tests/isolate_hook.py
```

The helper:

1. Snapshots the vendored `_compare` source bytes before modification
2. Applies a patch to isolate a single hook
3. Invalidates cached bytecode by default so subprocess imports observe the patched source
4. Restores the original source on exit
5. Verifies restoration using SHA-256 checksums
6. Raises loudly if the vendored submodule would be left in a modified ("dirty") state

This guarantees byte-for-byte restoration after every test execution.

#### Backward Compatibility

* `patch_compare_model` has been retained as a delegating alias to `isolated_hook`

#### Test Migration

Updated `test_per_hook_isolation.py` to use `isolated_hook`.

Added a new test:

```text
test_dirty_restore_raises_loudly
```

which validates the failure path when restoration cannot return the source tree to its original state.

---

### Source-Patch Isolation Tests

* ✅ Per-hook isolation suite: **15 tests passed**
* ✅ Includes coverage for the new dirty-restore detection path

### Build & Collection Checks

* ✅ All modified Python files compile successfully
* ✅ Both vLLM E2E test suites collect without errors
* ✅ All three shell scripts pass:

```bash
bash -n <script>
```

### DIFF
--- a/tests/isolate_hook.py
+++ b/tests/isolate_hook.py
@@ -13,8 +13,11 @@ disabled; otherwise the captured tensor for hook H is correct but the
 forward pass also writes to other buffers, which keeps the un-hooked
 tensors live and could mask hook-perturbation regressions.
 
-This module owns the source-level patching.  It is callable both as a
-library (``patch_compare_model``) and as a CLI for ad-hoc isolation runs.
+This module owns the source-level patching.  The canonical entry point is
+the ``isolated_hook`` context manager, which snapshots the file bytes,
+patches, and on exit restores *and asserts byte-identical restoration* so
+the vendored submodule is never left dirty.  It is also callable as a CLI
+for ad-hoc isolation runs.
 
 The patch is applied **in-place** to the source file, with the original
 saved to a sibling ``.copy_isolate_backup`` so a Ctrl-C / crash doesn't
@@ -30,6 +33,7 @@ from __future__ import annotations
 
 import argparse
 import contextlib
+import hashlib
 import os
 import re
 import shutil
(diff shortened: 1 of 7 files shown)


Samfisheryu:
Heads up: this PR is still stacked on `test/marker-taxonomy-categorize` instead of `main`, and it was branched from an older #56 commit before the later CPU-marker fixes landed. Since #56 was squash-merged into `main`, retargeting this PR directly to `main` creates conflicts in the files touched by those later #56 fixes.

I can rebase and merge this one, but for future PRs let us avoid this nested/stacked branch pattern unless we explicitly agree on a stack. One issue should generally map to one PR based on `main`; otherwise reviews and squash merges make the following PRs harder to review and merge cleanly.

PR #58, opened 2026-06-11
(#3) test: shared E2E lib + configurable matrix

SieDeta:
# Summary

This PR introduces a reusable E2E comparison library and a configurable E2E matrix harness to eliminate duplicated comparator logic, improve observability, and provide robust matrix execution with per-cell isolation.

The changes are intentionally additive, preserving existing runners and inference paths while consolidating shared functionality.

### Shared E2E Comparison Library

Consolidates logic that was previously reimplemented across four separate comparators.

#### `compare.py`

Provides a unified comparison interface supporting:

* `bitwise`
* `allclose`
* `row_count`
* `transport_bitwise`

Also supports `Check` dataclass comparisons.

Additional diagnostics are recorded even for passing comparisons:

* Maximum absolute difference (`max_abs_diff`)
* Mean absolute difference (`mean_abs_diff`)
* First-difference position

This provides visibility into "barely passing" comparisons and enables future trend analysis.

---

#### `align.py`

Centralizes alignment logic including:

* Left-padding stripping
* EOS token trimming
* Request ID parsing (`<group_id>:<row_id>`)
* vLLM UUID suffix normalization
* Logits head-skip alignment

---

#### `report.py`

Standardizes reporting functionality:

* `CellResult` → JSON record conversion
* JSONL artifact generation
* Human-readable summary tables
* Legacy `result.json` compatibility adapter

---

#### `clickhouse_io.py`

Introduces shared ClickHouse utilities:

* Canonical dtype mappings
* Hook → ClickHouse `act_name` mappings
* Row decoding utilities
* Per-hook count aggregation
* Request grouping logic

Additional improvements:

* `clickhouse_driver` is imported lazily to preserve CPU importability.

---

#### Supporting Modules

* `segments.py`

  * Thin wrapper around `monitoring.segment_merger`

* `disk_ref.py`

  * Thin wrapper around `.pt` reference file parsing

* `hf_reference.py`

  * Thin wrapper around HF ROL/GEN rollout implementations

> **Note:** The larger (~700 LOC) module migration is deferred to the future legacy-removal PR to keep this PR additive.

### Matrix Expansion

Supports multi-value axes that expand into a Cartesian product of execution cells.

Each generated cell dispatches to the existing runners and comparators as subprocesses. No inference logic is reimplemented.

---

### Environment Variable Translation

Maintains the environment contract introduced in PR #57 :

```text
E2E_HOOK_SELECTION → DMX_HOOK_SELECTION
```

Translation occurs independently for each subprocess.

---

### Per-Cell Isolation

Each matrix cell executes independently with:

* Dedicated subprocess environments
* Independent result records
* Failure isolation from neighboring cells

---

### Timeout Protection

Introduces `--cell-timeout`.

If a runner hangs:

* The affected cell is marked as `ERROR`
* JSONL artifacts remain valid
* Summary tables are still generated
* Matrix execution continues without blocking indefinitely

---

### Dry-Run Mode

Introduces `--dry-run`, which:

* Expands and prints all planned matrix cells
* Displays dispatch commands
* Validates environment translation
* Requires no CUDA or ClickHouse dependencies

This enables CPU-only testing of matrix expansion behavior.

## Verification

* ✅ Added **37 new CPU-marked unit tests** (`tests/test_e2e_lib.py`)
* ✅ All **37/37 unit tests pass**
* ✅ Full CPU suite: **108 passed** (up from **71** after PR #57 )
* ✅ Real-run smoke tests confirm hung runners are recorded as `ERROR`
* ✅ Valid JSONL artifacts and summary tables are generated after failures
* ✅ Dry-run mode validated across HF/vLLM × comparison standards


tests/e2e_matrix.py
```diff
@@ -0,0 +1,398 @@
...
+            config_file = os.path.join(ref_dir, "ref_config.json")
+            steps.append(Step("enable_ref_hooks", ["<in-process>", "enable_ref_hooks"]))
+            steps.append(Step("vllm_ref", _runner("tests.vllm_ref_runner", "--output-dir", ref_dir)))
+            steps.append(Step("vllm_monitored", _runner("tests.vllm_monitored_runner", "--output-dir", mon_dir)))
+            steps.append(Step("compare", _runner(
+                "tests.vllm_identical_comparator",
+                "--ref-config", config_file, "--mon-dir", mon_dir,
+                "--result-file", result_file)))
+            return steps, "tests.vllm_identical_comparator", result_file
+        # row_count / allclose -> monitored-only + rowcnt comparator
```
Samfisheryu: `allclose` is advertised as a matrix standard, but this branch routes it to the row-count comparator. In `run_cell()`, the same non-identical vLLM path also writes `ref/meta.json` with `{"skipped": true}`, so `vllm_rowcnt_comparator` skips value comparison and `vllm + allclose` behaves like row-count-only. Please either reject/document this unsupported combination or wire an actual allclose comparison path.

tests/e2e_matrix.py
```diff
@@ -0,0 +1,398 @@
...
+        steps.append(Step("vllm_monitored", _runner("tests.vllm_monitored_runner", "--output-dir", mon_dir)))
+        steps.append(Step("compare", _runner(
+            "tests.vllm_rowcnt_comparator",
+            "--ref-dir", ref_dir, "--mon-dir", mon_dir,
+            "--result-file", result_file)))
+        return steps, "tests.vllm_rowcnt_comparator", result_file
+
+    if cell.backend == "hf":
```
Samfisheryu, line 198: HF ignores `cell.standard` here: `bitwise`, `row_count`, and `allclose` all run the same `hf_comparator`, whose behavior is the existing `max_abs <= E2E_TOLERANCE` tolerance check. That makes the advertised matrix standard misleading for HF. Please either limit HF to the standard that is actually implemented, or make the comparator path standard-specific.

Samfisheryu:
At the matrix level, the advertised `standard` axis does not currently match the actual test/comparator implementation.

`standard` is mostly used as a coarse comparator selector in `e2e_matrix.py`; it is not passed to the comparators (`--standard` / `E2E_STANDARD` do not exist), and each comparator still uses its hard-coded behavior. As a result, some matrix cells claim standards that are not actually tested: for example, `vllm + allclose` behaves like row-count-only, while HF `bitwise` / `row_count` still run the same tolerance comparator.

I do not think this PR should merge as-is. Please either fill in the missing test/comparator behavior so the advertised matrix standards are real, or restrict/validate the supported backend+standard combinations so the matrix cannot report misleading cells.

tests/test_e2e_tp2.py
```diff
@@ -0,0 +1,83 @@
...
+
+@pytest.mark.hf
+@require_gpus(2)
+@require_clickhouse()
+def test_e2e_correctness_hf_tp2(subtests) -> None:
+    """HF TP=2 eager: hooked model (ring -> ClickHouse) vs original model.
+
+    Equivalent matrix cell: ``--backend hf --mode eager --standard allclose --tp 2``.
+    """
+    argv = matrix_argv_from_env("hf", "allclose", mode="eager", env=_tp2_env())
```
Samfisheryu: This block presents the HF test as TP=2 coverage: `_tp2_env()` sets `E2E_TP_SIZE=2`, the test requires 2 GPUs, and the docstring/cell says `--tp 2`. However, the HF matrix path invoked here goes through `hf_reference_runner` and `hf_monitored_runner`, and those runners do not read `E2E_TP_SIZE`, do not launch under `torchrun`, and do not pass `tp_plan="auto"`. So this appears to require 2 GPUs without actually exercising HF tensor parallelism. Please either route this through a real HF TP execution path or remove/rename this test so it does not claim HF TP=2 coverage.

tests/numeric_study.py
```diff
@@ -0,0 +1,742 @@
...
+        sys.executable, "-c", runner,
+        "--model-key", model_key,
+        "--hook", hook,
+        "--mode", mode,
+        "--variant", variant,
+        "--rollout", rollout,
+        "--max-new-tokens", str(max_new_tokens),
+        "--prompt", prompt,
+        "--out", str(out_path),
+    ]
```
Samfisheryu, line 634: `_run_rollout()` also does not pass dtype to either child runner. Combined with the CLI `--dtype` argument below and the `StudyResult(dtype=...)` field, this makes dtype metadata-only today. Please include dtype in the subprocess contract if it is meant to be a real axis, or drop it from the CLI/report for now.

tests/numeric_study.py
```diff
@@ -0,0 +1,742 @@
...
+
+    MODEL_ALIASES = {
+        'gpt2': 'gpt2',
+        'qwen3': 'Qwen/Qwen3-0.6B',
+        'qwen2_moe': 'Qwen/Qwen1.5-MoE-A2.7B',
+        'llama': 'meta-llama/Llama-3.1-8B',
+    }
+    hf_id = MODEL_ALIASES[args.model_key]
+    device = torch.device('cuda')
+    dtype = torch.float16
```
Samfisheryu: The child HF runner does not parse a dtype argument and hardcodes `dtype = torch.float16`. This means the public `--dtype` option can change the recorded result/threshold selection without changing the actual HF rollout dtype. Please either wire dtype into the child runner or remove/limit the public dtype option until it is real.

tests/numeric_study.py
```diff
@@ -0,0 +1,742 @@
...
+    # at a large negative floor (sufficient for chosen-token + top-k drift).
+    vocab = int(getattr(llm.llm_engine.model_config.hf_config, 'vocab_size', 0)) or 1
+    N = len(ids)
+    logits = torch.full((N, vocab), -1e30, dtype=torch.float32)
+    for i in range(N):
+        if i < len(step_logprobs) and step_logprobs[i] is not None:
+            for tid, lp in step_logprobs[i].items():
+                logits[i, int(tid)] = float(lp.logprob)
+    token_ids = torch.tensor(ids, dtype=torch.int64)
+    torch.save({'token_ids': token_ids, 'logits': logits}, args.out)
```
Samfisheryu, line 584: This implementation confirms the vLLM path is top-k logprobs, not full-vocab logits: `logprobs=20` returns sparse per-step logprob dicts, then the code creates a dense-shaped tensor and fills unreported entries with `-1e30`. That can be acceptable if vLLM only exposes top-k here, but then metrics like vocab-wide `mean_abs` and any “full logits” wording are misleading. Please either report this as top-k logprob drift or restrict the metrics to values that are meaningful for sparse top-k capture.

tests/numeric_study.py
```diff
@@ -0,0 +1,742 @@
...
+                 + (f", {n_alert} alerting" if n_alert else ""))
+    return "\n".join(lines)
+
+
+# ---------------------------------------------------------------------------
+# Subprocess rollout runners (full-logits capture)
+# ---------------------------------------------------------------------------
+
+# Each rollout subprocess saves {token_ids: int64[N], logits: float32[N, vocab]}.
+# Same on-disk shape for HF and vLLM so the comparison code is framework-agnostic.
```
Samfisheryu: This section describes the rollout artifact as full `[N, vocab]` logits for both HF and vLLM. That is not true for the vLLM runner below, which requests top-k logprobs and fills missing vocab entries with `-1e30`. Please split the artifact semantics by framework or describe the vLLM artifact as sparse/top-k logprob based rather than full logits.

Samfisheryu (changes requested):
I merged the contents of #59 and #60 into this PR and closed those two PRs, so we can review the shared E2E matrix, wrapper migration, and numeric study together here.

The main things to fix before this can merge are:

1. Align the `standard` axis with the actual comparator behavior. Right now some standards are routed to comparators that do not actually implement that standard, especially `allclose`/`row_count` on the vLLM side and non-allclose standards on the HF side.

2. Align the matrix with the test implementation. If a combination is listed in the matrix, it should either be genuinely implemented and tested, or explicitly rejected/skipped with the right semantics.

I also left inline comments for the issues introduced by the merged #59/#60 content:
- the HF TP=2 test currently requires 2 GPUs but does not appear to actually run HF tensor parallelism;
- the vLLM numeric-study path captures top-k logprobs, not full logits, so the wording/metrics should reflect that;
- `--dtype` is currently metadata-only and should either be wired into the child runners or removed/limited.

This PR is not ready to merge yet. Please address the issues above first.

tests/e2e_matrix.py
SieDeta: Fixed. allclose is no longer a valid vLLM standard in plan_cell(), so now the code now raises ValueError at plan time with a message listing the supported standards (bitwise, transport_bitwise, row_count). The matrix cannot produce a vLLM+allclose cell; it will fail  before any subprocess is launched. The module docstring documents the supported combinations.

tests/e2e_matrix.py
SieDeta: Fixed. HF now accepts only allclose with any other standard raises ValueError at plan time. hf_comparator also receives --standard so it acts on it rather than ignoring it. The matrix cannot claim a HF+bitwise or HF+row_count test; those combinations are rejected before any subprocess runs.

tests/test_e2e_tp2.py
SieDeta: Fixed. test_e2e_correctness_hf_tp2 has been removed. The module docstring now explicitly states that HF TP=2 is not covered here because hf_reference_runner/hf_monitored_runner do not read E2E_TP_SIZE, do not launch under torchrun, and do not pass tp_plan="auto", so a 2-GPU HF test would be misleading. Only test_vllm_identical_tp2 remains, which genuinely exercises TP=2 sharding.

tests/numeric_study.py
SieDeta: Fixed. _run_rollout() now accepts dtype as an explicit parameter and passes --dtype to the child subprocess. dtype is a real execution axis, not metadata-only.

tests/numeric_study.py
SieDeta: Fixed. The child HF runner now parses --dtype and resolves it through _DTYPE_MAP = {'float16': torch.float16, 'bfloat16': torch.bfloat16, 'float32': torch.float32}. The public --dtype option now controls the actual rollout dtype end-to-end.

tests/numeric_study.py
SieDeta: Fixed. The section comment now splits artifact semantics by framework: HF saves full [N, vocab] logits; vLLM saves the same shape but only top-k logprob entries are real, with non-reported positions filled to -1e30. The module docstring states this upfront. compute_drift() uses sparse_floor=-1e30 to mask the vLLM tensor so all drift metrics (max_abs, mean_abs, top-k vocab diffs) are restricted to entries where at least one runner reported a real logit.

PR #63, opened 2026-06-15
fix(ring): comment out dead per-step host read of actual_bytes_counter (~430µs/step)

### LINKED ISSUE #62: [bug] ~6× per-hook monitoring overhead from a discarded managed-counter read in prepare_step (#51)
# `prepare_step` host reads a GPU-resident managed counter every step — per-step overhead added in #51, value discarded

## Summary

Builds that include #51 (`actual_bytes_counter`) show a per-step steady-state
overhead during monitoring that pre-#51 builds do not. The overhead is **fixed
per decode step** (engages once producers write to the ring, then saturates —
**not** linear in hook count) and measures ~0.5 ms/step on Llama-8B. A block in
`prepare_step` added by #51 **host-reads `actual_bytes_counter` (a
`cudaMallocManaged`, GPU-preferred counter the producer writes every step) and
then discards the result** (`(void)counter_delta`); it is absent pre-#51.

**Confirmed root cause:** commenting out the four-line read recovers the entire
regression (16.60 → 2.81 µs/hook, ≈ the pre-#51 2.48) — before/after below. The
removal is behavior-neutral (`counter_delta` is never consumed).

## Measured regression

Controlled head-to-head — **same node** (`gpu-a6-3`), same session, same fork,
same script, both `.so` built with identical flags, **shared baseline 6.2477
ms/step**, Llama-8B (32 hidden-states hooks):

| build    | 0 hooks | 16 hooks | 32 hooks | avg slope     |
|----------|--------:|---------:|---------:|---------------|
| pre-#51  |  6.2944 |   6.3361 |   6.3844 |  2.81 µs/hook |
| with #51 |  6.3109 |   6.8214 |   6.8636 | 17.27 µs/hook |

Segment analysis (with-#51) shows it is **not** per-hook:

- `0 → 16 hooks`: (6.8214 − 6.3109)/16 = **31.9 µs/hook** — a ~0.51 ms/step jump
- `16 → 32 hooks`: (6.8636 − 6.8214)/16 = **2.6 µs/hook** — flat, ≈ pre-#51's 2.81

i.e. a fixed ~0.5 ms/step cost that turns on once producers write data and then
saturates regardless of how many more hooks are added. Cross-job runs (separate
jobs, 3 models) agree:

| model      | pre-#51 µs/hook | with-#51 µs/hook | ratio |
|------------|----------------:|-----------------:|------:|
| Qwen3-4B   | 2.7             | 15.0             | 5.6×  |
| Llama-8B   | 3.4             | 16.7             | 4.9×  |
| Qwen3-14B  | 3.1             | 11.8             | 3.8×  |

## The code (introduced in #51 — `5792d6c`)

**`monitoring/csrc/ring/ring_engine_py.cu`** — top of `prepare_step`, runs every
call:

```cpp
// Read the device-side actual_bytes counter and compute the delta vs the
// snapshot from the previous prepare_step call. ... No cudaStreamSynchronize
// is required.
const uint64_t counter_cur = *impl_->engine.ring_state().actual_bytes_counter;   // host read
const uint64_t counter_delta = counter_cur - impl_->last_counter_read;
impl_->last_counter_read = counter_cur;
(void)counter_delta;  // consumed by reservation accounting in a follow-up
```

**`monitoring/csrc/ring/ring_alloc.h`** — the counter is unified memory, advised
and prefetched to the GPU:

```cpp
cudaMallocManaged(pp, sizeof(uint64_t));                       // managed
cudaMemAdvise(ptr, 8, cudaMemAdviseSetPreferredLocation, gpu_loc);   // home = GPU
cudaMemPrefetchAsync(state_.actual_bytes_counter, 8, gpu_loc, 0, stream);  // -> GPU
```

**`monitoring/csrc/ring/producer.cu:147`** — the producer writes it on the
compute stream every time it fires:

```cpp
atomicAdd(reinterpret_cast<unsigned long long*>(ring.actual_bytes_counter), ...);
```

## Confirmed (before / after, one change)

Three builds in one job, same node, same session, **shared baseline 6.2721
ms/step**. The only change between rows 2 and 3 is commenting out
`ring_engine_py.cu:196-199`:

| build                        | slope        | full overhead |
|------------------------------|-------------:|--------------:|
| no counter (pre-#51)         | 2.48 µs/hook |        +2.07% |
| with counter (current)       | 16.60 µs/hook|        +9.40% |
| counter read removed         | 2.81 µs/hook |        +2.58% |

Removing the read drops the slope 16.60 → 2.81 µs/hook — back to the no-counter
2.48 within noise — recovering **~430 µs/step** (matches a fine-sweep's ~450 µs
plateau). The read accounts for essentially the entire regression.

## Suggested direction

The reclamation feature the counter was added for is legitimate (upper-bound
reservations for variable-byte / EP producers leak ring space without it). The
issue is only the per-step host read on the critical path while the consumer
does not exist. Options: remove the read until the reclamation consumer lands;
when it lands, read the counter off the critical path (e.g. via the drain
thread, which already synchronizes) or via a CPU-side mirror, rather than a
host read of GPU-preferred managed memory inside `prepare_step`.

zhaoyboo:
The `actual_bytes_counter` read in `prepare_step` was added for **chunked-producer
  ring reclamation** (variable-byte / EP producers that reserve an upper bound and
  need the actual bytes to reclaim the difference).

  **No hook uses the chunked producer today** — the vLLM adapter wires only the
  prefix and basic producers — so the value is unused (`(void)counter_delta`).

  It still costs: the read is a host access of a GPU-resident `cudaMallocManaged`
  page the producer writes every step → a UVM coherence stall, **~430 µs/step on
  Llama-8B** (per-enabled-hook slope 16.6 → 2.8 µs/hook when removed; behavior-neutral).

  This PR comments it out until the chunked producer is actually used; when it is,
  re-add the read **off** the `prepare_step` critical path (e.g. on the drain thread).

### DIFF
--- a/monitoring/csrc/ring/ring_engine_py.cu
+++ b/monitoring/csrc/ring/ring_engine_py.cu
@@ -186,17 +186,30 @@ int RingEnginePy::prepare_step(uint64_t step_total_bytes,
 {
     impl_->current_hook_idx = 0;
 
-    // Read the device-side actual_bytes counter and compute the delta vs the
-    // snapshot from the previous prepare_step call.  The counter is monotonic
-    // and atomically updated inside the producer kernel's last-block-arrives
-    // section; the existing __threadfence() in producer.cu orders the D2D
-    // stores before the increment, so any delta we observe corresponds to
-    // bytes whose write is already globally visible.  No cudaStreamSynchronize
-    // is required.
-    const uint64_t counter_cur = *impl_->engine.ring_state().actual_bytes_counter;
-    const uint64_t counter_delta = counter_cur - impl_->last_counter_read;
-    impl_->last_counter_read = counter_cur;
-    (void)counter_delta;  // consumed by reservation accounting in a follow-up
+    // actual_bytes_counter reclamation: DISABLED for now (see below).
+    //
+    // The counter exists to reclaim ring space when a step's reservation
+    // OVER-estimates what the producer actually writes.  That only happens for
+    // producers whose written byte count the CPU cannot size up front -- i.e.
+    // variable-byte / EP "chunked" producers that reserve an upper bound.  No
+    // hook currently uses that path: the vLLM adapter only wires the prefix
+    // producer (CPU-known actual_q_len * row_bytes) and the basic producer
+    // (CPU-known x.nbytes()), both of which reserve exactly what they write and
+    // need no reclamation.  The reclamation consumer was also never landed, so
+    // the delta is unused.
+    //
+    // Reading the counter here is NOT free: it is a host dereference of a
+    // cudaMallocManaged page whose preferred location is the GPU and which the
+    // producer writes every step, so the read forces a UVM coherence stall
+    // (measured ~430 us/step on Llama-8B -- effectively a per-step implicit GPU
+    // sync, despite no explicit cudaStreamSynchronize).  Keep it commented out
+    // until a chunked-style producer AND a reclamation consumer actually exist;
+    // when they do, read the counter OFF the prepare_step critical path (e.g.
+    // on the drain thread, which already synchronizes) rather than here.
+    //
+    // const uint64_t counter_cur = *impl_->engine.ring_state().actual_bytes_counter;
+    // const uint64_t counter_delta = counter_cur - impl_->last_counter_read;
+    // impl_->last_counter_read = counter_cur;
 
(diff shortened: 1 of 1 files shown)


XbzOnGit (approved):
LGTM

PR #64, opened 2026-06-16
[Feature] get_internal: read captured internals back as HF-style tuples

Samfisheryu:
## What this adds

A read-side API that hands DMI's captured internals back in the **same shape HuggingFace gives you**, so consuming them is as familiar as `out.hidden_states`.

**Before** — off the low-level reader you reassemble everything by hand, and repeat it for every internal:

```python
rows = reader.prefix_get((model_id,))
by_layer = {}
for (mid, req, act, layer, shard, s, e), t in rows:
    if act == "blocks.hook_resid_pre":
        by_layer.setdefault(layer, []).append((s, t))
# concat each layer's chunks by token, then left-pad the ragged requests into a
# batch, then stack into [layer, batch, seq, hidden] ... and do it all again for
# attention, logits, kv, each with its own quirks
hidden = torch.stack([...])   # many lines, easy to get wrong
```

**After**:

```python
from integration.hf_adapter import generate_with_monitoring
from monitoring.internal_mapper import get_internal

out = generate_with_monitoring(model, **inputs, max_new_tokens=8,
                               hook_selection="resid_pre")

internal = get_internal(out)
internal.available        # ['hidden_states'] — what this run actually captured
internal.hidden_states    # tuple indexed by layer, each [batch, seq, hidden]
                          # exactly like HF's out.hidden_states
```

`get_internal` also takes a bare `model_id` + `reader` for cross-process / after-the-fact analysis (no model, no GPU):

```python
internal = get_internal("my_run_id", reader)
```

## Core changes — what & why

- **`monitoring/internal_mapper.py`** (new) — backend-agnostic read layer. Reads the per-`(layer, token-range)` rows DMI writes, groups by layer, concatenates along the token axis, and **left-pads ragged requests** into `[batch, seq, hidden]` (same layout as HF's left-padded batch — on CPU, after the run, no effect on generation). Returns an `Internal` whose fields mirror an HF model output; `.available` lists captured fields, and accessing an uncaptured one raises with a hint. **Extensible by design**: a `field -> (act_name, reassembler)` table (`_FIELDS`) — adding `attention` / `logits` / `kv` later is one entry + its reassembler, with no change to `get_internal`.
- **`integration/hf_adapter.py`** — `generate_with_monitoring` now returns `MonitoredGenerateOutput(.sequences, .model_id)` instead of a bare tensor. `.model_id` is the handle `get_internal` uses; any other attribute (`scores`, `logits`, tensor methods…) **falls through to the underlying HF output**, so existing call sites keep working.

## Incidental changes

- **`example/visualization/run_offload_hf.py`** — `output_ids[0]` → `output_ids.sequences[0]` for the new return type.
- Other `generate_with_monitoring` callers (`hf_compare_runner`, `test_per_hook_isolation`, `test_e2e_correctness_vs_hf`, benchmarks) unchanged — they rely on attribute fall-through.

## Testing

- **Unit** (`tests/test_internal_mapper.py`) — 5 passed: per-layer reassembly, chunk-concat order, ragged left-pad, uncaptured-field error, `source` accepts `out` or `model_id`.
- **E2E** (`tests/test_internal_mapper_e2e.py`, Qwen3-0.6B: CUDA + ClickHouse + fork) — 8 passed: return type, **output identical to native HF generate**, `.available`, tuple shape / layer count / hidden dim, token count (drops the unforwarded last token), `model_id`+`reader` path, uncaptured-field error, ragged-batch left-pad. Each run uses a uuid `model_id` with teardown cleanup — 0 leftover rows.

### DIFF
--- a/docs/huggingface.md
+++ b/docs/huggingface.md
@@ -44,6 +44,7 @@ requirements = InternalRequirements().require(
     count=model.config.num_hidden_layers,
     retry=True,
     timeout_s=30.0,
+    match_token_ranges=True,
 )
 
 out = generate_with_monitoring_dict(
@@ -60,7 +61,9 @@ token_mask = out.dmi_internal.token_mask
 
 For layer-tuple fields such as `hidden_states`, `count` validates the number of
 layers in the reassembled tuple. It does not validate token completeness inside
-each layer tensor.
+each layer tensor unless `match_token_ranges=True` is set. That option checks
+the captured row ranges against the token ranges recorded during this generate
+call. For per-layer fields it uses a fast representative-layer check.
 
 Supported mapped fields are:
 
@@ -119,8 +122,9 @@ norm = hidden_states[0].float().norm(dim=-1)[token_mask].mean()
 
 If a tensor field and `token_mask` have different `[batch, seq]` shapes, it
 usually means the tensor field was read before all rows for that field arrived.
-Current requirements for layer-tuple fields validate layer count, not token
-count, so clear that field's cache and read it again:
+Use `match_token_ranges=True` with `retry=True` to wait for expected token
+ranges. If you already cached a partial field, clear that field's cache and read
+it again:
 
 ```python
 out.dmi_internal.clear_cache("hidden_states")
@@ -159,6 +163,7 @@ out.dmi_internal.require(
     retry=True,
     timeout_s=30.0,
     poll_s=0.25,
+    match_token_ranges=True,
 )
 hidden_states = out.dmi_internal.hidden_states
 ```
--- a/docs/huggingface.md
+++ b/docs/huggingface.md
@@ -42,6 +42,8 @@ from monitoring.internal_mapper import InternalRequirements
 requirements = InternalRequirements().require(
     "hidden_states",
     count=model.config.num_hidden_layers,
+    retry=True,
+    timeout_s=30.0,
 )
 
 out = generate_with_monitoring_dict(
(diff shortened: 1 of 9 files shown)


XbzOnGit:
Update on the HF-facing API:

`generate_with_monitoring(...)` now keeps normal HF return behavior. Users who
want DMI internals should use `generate_with_monitoring_dict(...)`, which always
returns a dict-style HF output and attaches one extension: `out.dmi_internal`.

```python
from integration.hf_adapter import generate_with_monitoring_dict
from monitoring.internal_mapper import InternalRequirements

requirements = InternalRequirements().require(
    "hidden_states",
    count=model.config.num_hidden_layers,
    retry=True,
    timeout_s=30.0,
    poll_s=0.25,
    match_token_ranges=True,
)

out = generate_with_monitoring_dict(
    model,
    **inputs,
    max_new_tokens=8,
    hook_selection="hidden-states",
    internal_requirements=requirements,
)

hidden_states = out.dmi_internal.hidden_states
token_mask = out.dmi_internal.token_mask
```

Incomplete data behavior:

- Plain access is lazy and caches successful reads.
- Missing fields, too few rows, or failed token-range checks raise
  `IncompleteInternalError`.
- `retry=True` retries until complete or `timeout_s` expires.
- `count` currently checks field-entry count, so for `hidden_states` it checks
  layer count.
- `match_token_ranges=True` checks captured row ranges against token ranges
  recorded during this `generate()` call.

Manual clear and re-read:

```python
from monitoring.internal_mapper import IncompleteInternalError

try:
    hidden_states = out.dmi_internal.hidden_states
except IncompleteInternalError:
    out.dmi_internal.clear_cache("hidden_states")
    hidden_states = out.dmi_internal.hidden_states
```

Future work:

- Incremental retry/reassembly instead of re-reading the whole field.
- Richer and more efficient requirement checks for token counts / token coverage.
- Metadata-only completeness checks before loading tensor payloads.
- Better missing-token detection using recorded token-range metadata.

XbzOnGit (approved):

Samfisheryu:
Good

PR #66, opened 2026-06-17
[feature] DMILLM: drop-in vLLM LLM with per-request .dmi_internal

Samfisheryu:
## What this adds

The vLLM counterpart to `get_internal` / `dmi_internal` (#64). Native vLLM returns text + token ids + logprobs and **no internals at all** — there is no `output_hidden_states` equivalent. `DMILLM` is a drop-in `LLM` subclass, so the call site stays native:

**Native vLLM:**
```python
from vllm import LLM, SamplingParams
llm = LLM("Qwen/Qwen3-0.6B")
out = llm.generate(["The capital of France is"], SamplingParams(max_tokens=8))
out[0].outputs[0].text   # that's all you get
```

**With DMI:**
```python
from vllm import SamplingParams
from integration.vllm_adapter import DMILLM      # ← LLM → DMILLM

llm = DMILLM("Qwen/Qwen3-0.6B",
             additional_config={"dmx_model_id": "demo_vllm",
                                "dmx_db_host": "localhost", "dmx_db_port": 9000})
out = llm.generate(["The capital of France is"], SamplingParams(max_tokens=8))
out[0].outputs[0].text             # native vLLM output, unchanged
out[0].dmi_internal.hidden_states  # per-layer tuple, each [1, seq, hidden]
```

`llm.generate` is 100% native; the only change is `LLM` → `DMILLM`. This mirrors the HF path's `out.dmi_internal`.

## Core changes

- **`integration/vllm_adapter.py`** — `DMILLM(LLM)`: `__init__` injects `worker_cls` and stores model_id / db connection; `generate` tags each `RequestOutput` with a lazy `.dmi_internal` keyed by its normalized `request_id`. Plus an `_attach_dmi_internal` helper. Reuses the existing `make_lazy_internal` and `normalize_vllm_request_id`.
- **Per-request**: each `RequestOutput` exposes only its own request's internals — `[1, seq, hidden]`, real length, no cross-padding (vLLM returns one `RequestOutput` per prompt). For the whole batch as one `[batch, seq, hidden]` tensor, `get_internal(model_id)` still works.
- DMI settings are passed through `additional_config` exactly as with plain `LLM`.

## Docs

- **`docs/vllm.md`** — new "Reading internals back: `DMILLM`" section.

## Testing

- **Unit** (`tests/test_vllm_dmillm.py`) — 3 passed: per-request tagging, isolation across prompts, reader passthrough (no GPU / ClickHouse).
- **E2E** (`tests/test_vllm_dmillm_e2e.py` + `tests/vllm_dmillm_runner.py`) — subprocess runner so the pytest parent never touches CUDA (per repo convention for vLLM tests). Starts a real vLLM engine on Qwen3-0.6B and reads each `RequestOutput`'s `.dmi_internal` back from ClickHouse. Checks:
  - native output unchanged;
  - **internals readable without an explicit `stop_monitoring`** — the drain-flush timeout lands captures in ClickHouse automatically, matching the post-`stop` baseline;
  - per-request `[1, seq, hidden]`;
  - ragged isolation (12 vs 8 tokens for `"The capital of France is"` vs `"Hello"`);
  - `available == ['hidden_states']`.
  
  **Passed.**

## Note

The e2e confirms internals are readable **without** an explicit flush (the drain-flush timeout auto-lands captures in ClickHouse). The runner still calls `stop_monitoring` once for a deterministic baseline; wiring auto-flush + lazy-retry into `DMILLM.generate` itself (so users never think about flush, even reading immediately after `generate`) is a follow-up PR.

### DIFF
--- a/docs/vllm.md
+++ b/docs/vllm.md
@@ -25,9 +25,6 @@ export VLLM_DISABLE_COMPILE_CACHE=1
 ## Offline API
 
 ```python
-import os
-os.environ["VLLM_DISABLE_COMPILE_CACHE"] = "1"  # required (see above)
-
 from vllm import LLM, SamplingParams
 
 llm = LLM(
@@ -60,9 +57,6 @@ the captured internals back from the store -- the same object the HuggingFace
 path's `out.dmi_internal` gives you.
 
 ```python
-import os
-os.environ["VLLM_DISABLE_COMPILE_CACHE"] = "1"  # required (see above)
-
 from vllm import SamplingParams
 from integration.vllm_adapter import DMILLM
 
--- a/docs/vllm.md
+++ b/docs/vllm.md
@@ -52,6 +52,45 @@ for o in llm.generate(["The answer is"], params):
 Set `"dmx_null_mode": False` and configure `dmx_db_*` fields to persist captures
 to ClickHouse.
 
+## Reading internals back: `DMILLM`
+
+`DMILLM` is a drop-in subclass of `LLM`: it injects the DMI worker for you, and
+every `RequestOutput` from `generate` carries a lazy `.dmi_internal` that reads
+the captured internals back from the store -- the same object the HuggingFace
+path's `out.dmi_internal` gives you.
+
+```python
+import os
+os.environ["VLLM_DISABLE_COMPILE_CACHE"] = "1"  # required (see above)
+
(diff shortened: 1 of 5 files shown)


XbzOnGit (approved):
LGTM

Issue #67, opened 2026-06-18
[BUG] vLLM continuous batching: captured activations attached to the wrong request (row↔request mismatch)


zhaoyboo:
# [BUG] vLLM continuous batching: captured activations attached to the wrong request (row↔request mismatch)

**Component:** `integration/vllm_adapter.py` (vLLM V1 capture path / step-context construction)
**Severity:** High — silent data-correctness bug. Captured tensors and metadata
look valid, so it passes smoke tests; only an accuracy/AUROC measurement exposes it.
**Status:** Root-caused and fixed (commit `07a69e0`, pure-Python, no native rebuild).
File this as the bug record / PR description.
**Affects:** any DMI consumer doing online capture under **continuous batching**
(≥2 concurrent requests). Single-request / batch-size-1 capture is unaffected.

---

## TL;DR

Under vLLM continuous batching, each per-request slice carries the **correct
metadata** (`req_id`, `start_token`, `end_token`) but the **wrong activation
value** — it gets another concurrent request's row from the batched
`[num_tokens, hidden]` forward. ~60% of requests are affected. Cause: the
row→request map was built (a) in `num_scheduled_tokens` **dict order** and (b)
**before** `model_runner._update_states`, while vLLM lays the rows out in the
**post-`_update_states` `input_batch.req_ids` order**.

---

## Symptom

A linear probe on `resid_pre` L28 (SEPs hallucination probe) trained offline
(HF `output_hidden_states`) achieves AUROC **0.811** held-out. Served online via
vLLM+DMI:

| Mode | AUROC |
|---|---|
| Single request per `generate` | **0.811** ✓ |
| All prompts in one `generate` (continuous batching) | **0.52** (≈ random) |

Same probe, same questions (deduped held-out), same token-selection logic
(`(req_id, end_token==prompt_len+N)`), targets present, 0 missing — only the
batching mode changed.

Direct check: for the **same question + same greedy answer**, the answer-final
L28 hidden captured single-request vs continuous-batch has **cosine median ≈
0.55** (~60% < 0.9). The value is wrong, not the metadata.

## Root cause

vLLM V1's model runner lays the flattened hidden-tensor rows out in
**`input_batch.req_ids` order**, established by `_update_states`
(`remove_request` / `add_request` / `condense` / `refresh_metadata`) and consumed
by `_prepare_inputs`:

```python
# vllm/v1/worker/gpu_model_runner.py (execute_model)
req_ids = self.input_batch.req_ids
[...]

PR #68, opened 2026-06-18
vllm: fix continuous-batch capture mis-associating activations with requests

zhaoyboo:
Under vLLM V1 continuous batching, each per-request capture slice carried the correct metadata (req_id/start_token/end_token) but the WRONG activation value: it got another concurrent request's row out of the batched [num_tokens, hidden] forward. ~60% of concurrent requests affected; single-request unaffected. Silent -- tensors and metadata look valid, only an accuracy/AUROC measurement exposes it.

Root cause (two parts):
- build_step_context ordered req_ids by scheduler_output.num_scheduled_tokens DICT order, but vLLM lays the flattened hidden rows out in input_batch.req_ids order (_prepare_inputs: np.repeat over num_scheduled in input_batch order).
- before_forward ran in execute_model BEFORE super().execute_model -> before the model_runner's _update_states (remove_request/add_request/condense), so it read the PREVIOUS step's input_batch order. Under continuous batching the two orderings diverge (freed-slot reuse/condense), so request i's metadata was attached to request j's row.

Fix:
- build_step_context orders req_ids by model_runner.input_batch.req_ids (the true row order), falling back to dict order when there is no input_batch.
- before_forward is driven from a _update_states wrapper installed in load_model, so the step context is built AFTER input_batch reflects this step and before the forward; removed the pre-super() call.

Pure-Python, no native rebuild. Verified (Qwen3-8B, held-out TriviaQA, exact- position selection): continuous-batch AUROC 0.52 -> 0.802; single-request 0.811 -> 0.811 (both == offline HF 0.811). A standalone reproducer (repro_continuous_batch_row_mismatch.py) goes BUG -> PASS across this change: token-matched cosine(single, continuous) mean 0.555 (81% < 0.9) -> 0.9998 (0%).

Refs #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,15 +118,16 @@ llm = LLM(
     worker_cls="integration.vllm_adapter.DMXGPUWorker",
     additional_config={
         "dmx_hook_selection": "vllm-full",
-        "dmx_null_mode": True,   # capture + transport, drop on host (no DB needed)
+        "dmx_null_mode": False,
+        "dmx_db_host": "",       # active capture + transport; no persistence
     },
 )
 
 for o in llm.generate(["The answer is"], SamplingParams(max_tokens=16)):
     print(o.outputs[0].text)
-# Internal states for every layer have been captured into Ring²
-# during the run. Set "dmx_null_mode": False and configure a sink
-# to persist them.
+# Internal states for every layer have traversed Ring² during the run.
+# Configure a sink to persist them. Setting "dmx_null_mode": True
+# disables DMI planning, metadata, and payload copying.
 ```
 
 | | |
--- a/benchmark/bench_vllm_transport.py
+++ b/benchmark/bench_vllm_transport.py
@@ -2,7 +2,8 @@
 
 Compares modes:
   baseline         -- plain vLLM generate, no monitoring
-  ring_null        -- ring transport active (GPU->ring->CPU), null sink (no DB write)
+  ring_null        -- public DMI no-op/null mode (no planning or payload copy)
+  ring_active      -- ring transport active, no persistence sink
   ring_db          -- ring transport + ClickHouse ingestion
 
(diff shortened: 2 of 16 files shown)


XbzOnGit:
This PR fixes a vLLM V1 continuous-batching bug where request metadata can be matched to another request's activation rows after packed request order changes.

It also handles the coupled eager/padding problem: DMI must decide whether to force eager before vLLM finalizes dispatch, while metadata and ring reservation must match the real final layout. Predicting one layout for both is unsafe, especially when TP/PP ranks have different hook sets.

The updated flow is:

1. Run a side-effect-free capacity preflight using the worst complete TP/PP rank-role plan, so every rank makes the same eager decision without extra communication.
2. After real input preparation, record the actual request order, token counts, ranges, and row offsets.
3. After real dispatch, build metadata and reserve the ring once using the actual layout and batch descriptor.

Missing or inconsistent state now fails fast instead of falling back to scheduler dictionary order.

Validation includes 115 focused CPU tests and GPU ground-truth tests for TP=1/PP=1, TP=2/PP=1, and TP=1/PP=2 with deliberate order divergence and CUDA graphs active. Performance A/B testing with Qwen3-0.6B measured 0–4.5% ring-active overhead across those configurations.

Samfisheryu:
Validated final HEAD 44f8433 with an end-to-end Qwen3-0.6B continuous-batching capture test (24 staggered requests, TP=1). For 408 token-matched single-request vs. continuous-batch resid_pre L14 captures: mean cosine = 0.9998, median = 0.9999, min = 0.9880, 0% below 0.9, with no duplicate capture keys. PASS — this closes the original request/activation row-mismatch repro.

Samfisheryu (approved):
Approved based on the end-to-end validation above.

PR #75, opened 2026-08-13
Add general model expansion for vLLM 0.27.1

zhaoyboo:
## Summary

Consolidates the general vLLM 0.27.1 model expansion into one review unit:

- Gemma 3 (experimental)
- Phi-3
- Mistral
- Falcon-H1
- LFM2
- dense Jamba2
- Granite
- OLMo 3
- Apertus
- ERNIE 4.5
- MiniCPM 4.1

It also centralizes reusable multimodal black-box input generation and model-shape normalization needed by the independent SOTA follow-up PRs.

## Commit pair

- DMI: `6b955f16df8d899d73120fd34652f1b9948bf1f3`
- vLLM-DMI: `db9e7cfccc0bab6e52b311f6e7259d74af5feaa7`
- vLLM-DMI branch: `dmi-v0.27.1-model-expansion`

## Validation

- Focused vLLM 0.27.1 CPU contract matrix: **386 passed**
- Python syntax validation for all changed root and vLLM-DMI files
- Commit-graph and diff checks for the nine independent SOTA follow-ups

GPU/H100 runtime qualification remains intentionally separate from this structural PR. Model-specific audit documents identify the runtime cases still requiring H100 evidence.

## PR organization

This PR supersedes the stacked model PRs #76 through #84. SOTA model support is split into parallel one-commit PRs based directly on this branch, and their shared H100 qualification matrix will follow separately.

AI assistance was used to reorganize the existing commits, audit interface dependencies, and run the compatibility tests.

PR #74, opened 2026-08-13
Support vLLM 0.27.1 V1 runner

zhaoyboo:
## Summary

- pin the integration gitlink to upstream vLLM `v0.27.1` plus DMI integration commit `fdfe631884ae318050ce371e472c1135f317cfa2` on branch `dmi-v0.27.1`
- port GPT-2, Qwen2/Qwen2.5, Qwen3, Llama, and Qwen2-MoE across target model, loader, runner, and MoE API drift
- adapt DMI worker request-layout and teardown behavior, including bounded EngineCore shutdown and distributed TCPStore-owner grace
- add agent-driven API-boundary inventory, deterministic generated API-only black-box tests, version-pinned release-matrix gates, and retained evidence
- document the exact supported cells and the next model-family roadmap

## Validated scope

The support claim is deliberately limited to the official `vllm==0.27.1` wheel, V1 offline `LLM.generate`, CUDA/BF16, eager and the tested default CUDA-graph path, and only the matrix topologies below.

GPT-2 and Qwen3 passed public TP1 plus storage TP1/TP2. Qwen2 passed public TP1. Llama-3.1-8B and Qwen1.5-MoE passed public/storage TP2. V2, serve/async, PP/DP/EP/SP, quantization, speculative decoding, multimodal inputs, and unlisted configuration branches remain unclaimed and fail closed where applicable.

The exact runtime evidence pair is DMI root `ed43791eedac99c6fb18e24af8253e780bd56a54` plus integration `fdfe631884ae318050ce371e472c1135f317cfa2`. PR HEAD `4ba354bc330c22e093ade321d873ff0cfb2bffbc` adds only the final evidence documentation after that matrix.

## Tests

- complete release matrix: 18/18 passed, 0 failed, 0 blocked
- focused compatibility/API contracts: 219/219 passed
- public API-only black-box matrix: 5 model cells x eager/graph = 10/10 passed
- ClickHouse storage matrix: 12/12 cells and 395,896/395,896 rows bitwise equal
- zero fatal runtime-log markers, no residual vLLM process, and a passing three-sample post-run idle check
- the four graph+TP2 storage cells retain one named inner-executor shutdown warning reproduced against the exact official wheel; frontend force-kills, Worker exceptions, and TCPStore races remain fatal

[...]

PR #79, opened 2026-08-13
Add LFM2 support for vLLM 0.27.1

zhaoyboo:
## Summary

- register Lfm2ForCausalLM as a bounded DMI model for vLLM 0.27.1
- add native architecture-neutral conv_in and conv_out hook families
- derive the effective aligned LFM2 MLP width and preserve heterogeneous manifest firing order
- add production public-API and tiny hybrid storage release cells, CPU contracts, and a frozen support audit

Depends on ProjectDMX/vllm-DMI#5.

## Validation

- focused CPU contracts: 273/273 passed
- official LiquidAI/LFM2.5-1.2B-Instruct black-box eager and CUDA Graph: 2/2 passed in 89.42 seconds
- tiny-random/lfm2 eager storage: 4160/4160 byte-exact
- tiny-random/lfm2 CUDA Graph storage: 4160/4160 byte-exact

## Supported cell

TP1, BF16, V1 offline LLM, standard Hugging Face weights, eager and default CUDA Graph. LFM2-MoE/VL, TP greater than 1, PP/DP/EP/SP, V2, serving, prefix caching, quantization, and speculative modes remain excluded.


zhaoyboo:
Superseded by #75, which consolidates the general vLLM 0.27.1 model expansion into one review unit. The old branch is retained for history.

XbzOnGit:
Thanks for this work. LFM2 was intentionally not carried into [DMI-vLLM-Integration #12](https://github.com/ProjectDMX/DMI-vLLM-Integration/pull/12) under the no-fork split from [DMI #74](https://github.com/ProjectDMX/DMI/pull/74); the original E2E evidence is not being rejected.

LFM2 requires distinct short-convolution input/output hook types, which the stable DMI integration API v1 does not expose. Mapping those tensors onto attention, MLP, residual, or SSM IDs would be semantically incorrect and could collide with a genuine hook at the same layer. Support should return in a follow-up that first adds a public convolution hook family to DMI core and its integration API, then ports the wrapper to the standalone repository.

PR #80, opened 2026-08-13
Add dense Jamba2 support for vLLM 0.27.1

zhaoyboo:
## Summary

- register Jamba2ForCausalLM as a bounded dense DMI model for vLLM 0.27.1
- add native Jamba Mamba hook families and heterogeneous manifest construction
- add official-checkpoint black-box and storage release cells, compiler-sensitive constructor regression coverage, CPU contracts, and a frozen support audit
- fail closed for every Jamba MoE configuration and unknown layer kind

Depends on ProjectDMX/vllm-DMI#6.

## Validation

- focused CPU and release contracts: 284/284 passed
- official ai21labs/AI21-Jamba2-3B black-box eager and CUDA Graph: 2/2 passed in 94.13 seconds
- official checkpoint eager storage: 37083/37083 byte-exact
- official checkpoint CUDA Graph storage: 36557/36557 byte-exact

## Supported cell

TP1, BF16, V1 offline LLM, standard Hugging Face weights, eager and default CUDA Graph, dense Jamba2 only. Jamba MoE, AI21-Jamba2-Mini, TP greater than 1, PP/DP/EP/SP, V2, serving, prefix caching, quantization, and speculative modes remain excluded.


zhaoyboo:
Superseded by #75, which consolidates the general vLLM 0.27.1 model expansion into one review unit. The old branch is retained for history.

XbzOnGit:
Thanks for this work. This model was intentionally not carried into [DMI-vLLM-Integration #12](https://github.com/ProjectDMX/DMI-vLLM-Integration/pull/12) under the no-fork split from [DMI #74](https://github.com/ProjectDMX/DMI/pull/74); the original E2E evidence is not being rejected.

The model requires distinct SSM input/output hook types, which the stable DMI integration API v1 does not expose. Mapping those tensors onto attention, MLP, or residual IDs would be semantically incorrect and could collide with a genuine hook at the same layer. Support should return in a follow-up that first adds a public SSM hook family to DMI core and its integration API, then ports this wrapper to the standalone repository.

PR #95, opened 2026-08-23
Refactor project into a clean src-based layout

zaoxing:
## Summary

- reorganize the Python package under `src/dmi/` with dedicated adapters, hooks, storage, transport, and API modules
- split the Hugging Face adapter and hook/transport responsibilities into focused modules
- rename legacy adaptor classes while preserving compatibility aliases
- move benchmarks, examples, documentation assets, native sources, and ring tests into the clean project layout
- update imports, packaging, build paths, tests, and documentation for the new structure

## Validation

- `python -m pytest -q`: 134 passed, 53 skipped
- `python -m compileall`: passed
- source distribution and native build-path checks: passed

PR #113, opened 2026-08-26
Capture storage: immutable packs through bounded analysis (Phases 1-5), Python Reference Implementation

zaoxing:
## What this is, plainly

When we watch a model think, we capture a lot of tensors very fast. The obvious way to store them is to write each one into the database as it arrives — but then capture runs only as fast as the database accepts writes, and a slow database slows down the thing we were trying to observe.

So this does it the other way around. Captures are packed into files, the file is sealed and written to object storage once, and nothing waits on a database. Afterwards, a separate step walks those files and builds a searchable catalog in ClickHouse describing what is in them. Nobody is blocked while that catalog is being built.

The payoff comes at read time. The catalog knows exactly where each tensor sits — which file, which byte offset, how long it is — so answering "give me these forty tensors" fetches only those byte ranges, not the files containing them. The files are large; the tensors you want usually are not.

This PR adds all of that: **Phases 1 through 5** of the plan in `docs/capture-storage-design.md`. None of it is on `main` today — the whole `src/dmi/storage/capture/` package is new, about 4,600 lines across 14 modules.

It is opt-in and host-only. There are no CUDA changes, the existing ClickHouse payload sink is still the default, and nothing outside `storage/capture/` imports any of this yet. Merging it changes no current behaviour.

## What each phase adds

**Phase 1 — the file format.** Defines what a capture pack looks like on disk: identifiers, checksums, and a footer listing everything inside it. The footer sits at a known place, so a reader can learn a file's whole contents with two small reads instead of downloading it. Tested against truncation, corruption, unknown versions, and repeated writes.

**Phase 2 — the host pipeline.** Takes captures as they arrive and turns them into packs, sealing a pack when it gets big enough or old enough. Memory and disk are bounded: there is a fixed-size queue with an explicit policy for what to drop when it fills, so sustained overload cannot grow without limit. Packs can go straight out, or through a durable local spool that survives a restart.

**Phase 3 — object storage.** Uploads packs to any S3-compatible store, with multipart uploads, safe retries, listing, and the two small range reads used to load a footer. Verified against a pinned Garage 2.3.0 release.

**Phase 4 — the catalog.** Reads the footer of each uploaded pack and writes a row per capture into ClickHouse. It only reads footers, never payloads, so cataloguing stays cheap no matter how large the tensors are. Duplicate notifications are collapsed before any work happens, and packs whose notification went missing are found by scanning the bucket.

**Phase 5 — search and summaries.** Phase 4 could write the catalog but not read it, so this adds the read side: searching the catalog, paging through results stably, fetching only the byte ranges you selected, and computing summary statistics over the tensors you fetched.

## The one correctness problem worth reading

ClickHouse's `FINAL` keyword is the normal way to collapse duplicate rows, and Phase 4's views use it. It cannot be used for a *point-in-time* read, and the reason is subtle enough to be worth stating.

Each catalog row carries a version. To read the catalog as it stood at version `W`, the obvious query is "collapse duplicates, then keep rows at or below `W`". But `FINAL` collapses to the *newest* version that exists, and only then applies the filter. So a capture that was re-catalogued after `W` collapses to its new version, gets filtered out, and **vanishes from the result entirely** — rather than falling back to the version that was current at `W`.

Measured on ClickHouse 26.9.1, asking for the catalog as of version 1, where one capture was re-catalogued at version 2:

```
FINAL, then filter  -> [('capture-b', 128)]                      # capture-a has disappeared
correct query       -> [('capture-a', 64), ('capture-b', 128)]
```

Reads therefore go to the underlying tables and pick the newest version at or below the watermark explicitly. The existing `FINAL` views are untouched and still fine for ad-hoc queries.

## Numbers

All measured on an Apple Silicon laptop against local services. These are regression baselines for the Python reference implementation — **not** production capacity claims, and every one should be repeated on real hardware before anyone plans against it.

| Phase | What was measured | Result |
|---|---|---:|
| 1 | Building packs, 2,000 × 64 KiB captures | 0.521 GiB/s, 0.9% size overhead |
| 2 | Pipeline, direct vs. durable spool | 0.328 / 0.345 GiB/s, nothing dropped |
| 3 | Upload to Garage, 64 MiB packs, 4 workers | 0.502 GiB/s |
| 4 | Catalog inserts, 50,000-row batches | 157,567 rows/s |
| 5 | Point-in-time read vs. plain `FINAL` read | 22.3 ms vs. 12.1 ms |
| 5 | One page of results, first page vs. 25th | 21.5 ms → 24.4 ms |

Two of these are easy to misread:

- The point-in-time read is about **1.85× slower** than the `FINAL` read, but they are not two ways of doing the same thing. As shown above, `FINAL` returns the wrong answer. The gap is what correctness costs, not a menu option.
- Page 25 costs about the same as page 1. That is the point of how paging is implemented here — cost depends on page size, not on how deep you have gone. A more common approach would get steadily slower the further you page.

## Testing

425 tests in the standard CPU suite, 20 tests against a real ClickHouse, and one against a real Garage.

Two commits fix gaps in what was being tested, rather than adding more tests:

**CI cannot run on Phase 4 as it stands.** Two old test files came back that import a package removed in #95. Pytest imports every test file before it applies marker filtering, so the failure happens during collection and aborts the whole run — a file nobody meant to run takes everything down with it:

```
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
58 deselected, 1 error in 0.94s
make: *** [test-cpu] Error 2
```

That is `make check`, which is the whole of the CI job, and its `test-cpu` step is the standard CPU command. To be clear about the scope: the CPU tests themselves are perfectly healthy — skip the two files and all 277 of them pass. The problem is that the default command exits non-zero having run nothing. Those files were already replaced under new names in #95; the stale copies are removed here.

**Nothing tested the two halves together.** The Garage tests never touched ClickHouse, and the ClickHouse tests used made-up catalog entries pointing at a file that does not exist — so nothing they returned could actually be fetched, and the component that reads footers and writes catalog rows was covered by no live test at all. The new end-to-end test runs the whole path on real data: real tensors, packed, uploaded, catalogued, searched, fetched, decoded, and compared against the originals. To check it was worth adding, I introduced a deliberate bug that records byte offsets four bytes off — the old tests all still passed; the new one failed five ways.

## Problems found and fixed

Four, all in code from earlier phases:

1. **One oversized capture could kill the pipeline permanently.** The queue checks a capture against the queue's size limit, which has nothing to do with the maximum pack size. So a capture too big for any pack was accepted, then rejected later on the background thread, where the error was treated as fatal — the pack being assembled was thrown away and the pipeline stopped for good. It is now rejected up front, and the remaining edge case is counted and skipped instead of being fatal.
2. **The upload integrity check could not fail.** Uploads hand the file to the S3 transfer manager, so the upload code never sees the bytes. It recorded a checksum taken from the file's own metadata, and the later verification compared the stored object against that same recorded value — so it was comparing a number to itself. This ran immediately before deleting the local durable copy. Uploads now compute the checksum from the bytes actually sent.
3. **A batch-size limit was reported as a broken pack.** Exceeding a limit on total batch size was recorded as a failure of whichever pack happened to trip it, and the loop then carried on and skipped every remaining pack — while still reporting success.
4. **The Garage test harness could hang.** The server's error output went to a pipe nobody read; once the pipe filled, the server blocked, with no message explaining why.
The CUDA ring tests had the same problem — they pointed at `monitoring/csrc`, moved to `native/csrc` in #95, so `make` there failed immediately. That suite has since been dropped from this branch entirely, so **the branch is now CPU-only**: no `.cu`, `.cuh`, or `native/` file is touched or added.

## Phase 6 groundwork

Two of the instruments Phase 6 requires are included here, since neither existed and both define what a future native writer must satisfy.

**Fault injection** (`tests/_faults.py`) wraps the object store, the ClickHouse client, and the pack sink. Faults are scripted rather than random — a schedule names which calls fail and how — so a failure reproduces exactly. Nine tests characterise the required behaviour: a short read is refused rather than silently truncating, an immutable key written twice converges rather than conflicting, a sink failure fails loudly and then refuses admission, an insert failure leaves the pack uncommitted and the batch replayable, a duplicated insert is absorbed by replay semantics, and one corrupt pack fails only itself.

**A conformance manifest** (`tests/tools/golden_workload.py`) produces the golden-workload comparison the phase asks for: one JSON document over a deterministic corpus covering every dtype, recording pack identity and checksum, per-capture payload and decoded-tensor digests, placement, and the full summary contract. Every value is language-neutral, so a native writer is conformant exactly when the same corpus yields the same document.

Generating that manifest found a real defect in the Phase 5 summarizer on its first run: `l2_norm` used `sqrt(sum(x**2))`, which overflows float64 for large-magnitude tensors and returns infinity where the true norm is finite. It now factors out the largest magnitude before squaring.

## What this does not do yet

Stated here rather than buried: Phase 3's production capacity target is still unverified, because no representative hardware or target rate has been supplied. The catalog assumes a single writer, since row versions come from that process's own clock. Paging tokens are checked for tampering but not cryptographically signed. Search filters are applied before duplicate rows are collapsed, which is safe only because catalog entries are derived from immutable files. Each search costs one extra round trip. Looking up captures by ID does not use the table's sort order. Summary statistics ignore non-finite values, and report how many there were separately. A selection is a single page — callers page explicitly.

The live tests need a running ClickHouse and Garage, so they do not run in CI as configured.


tests/tools/run_garage_live.py
```diff
@@ -0,0 +1,161 @@
...
+            }
+        )
+        process = subprocess.Popen(
+            [binary, "server", "--single-node", "--default-bucket"],
+            cwd=root,
+            env=environment,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
```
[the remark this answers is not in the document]
zaoxing: Fixed in da4b931. Server stderr now goes to `garage.log` in the instance root rather than a pipe, so there is no buffer to fill; the handle is closed before termination. The tail of that log is also included in the error when the server exits during startup, which the PIPE version could never surface.

Good catch — this would only have shown up as a mid-run hang on a long benchmark, with no diagnostic.

src/dmi/storage/capture/catalog.py
```diff
@@ -0,0 +1,288 @@
...
+                        pack_id=ref.pack_id,
+                        object_key=ref.object_key,
+                        error_type=type(exc).__name__,
+                        message=str(exc)[:512],
+                    )
+                )
+                continue
+            estimated_bytes += pack_bytes
+            descriptors.extend(pack_descriptors)
+            valid_refs.append(ref)
```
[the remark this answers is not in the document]
zaoxing: Fixed in da4b931, exactly as suggested: `pack_bytes` is still computed inside the `try`, and the batch limit is enforced after the `except ... continue`, so it propagates instead of being swallowed.

The consequence you traced was the significant part — it was not just mislabelling. Because the loop continued, every remaining pack in the batch was skipped while `index()` still returned normally, so `CatalogReconciler.rebuild` reported success over an incomplete catalog.

Covered by `test_an_oversized_batch_raises_instead_of_blaming_a_pack`, which also asserts no descriptor or pack rows were written — the failure is atomic rather than partial.

tests/ring/Makefile
```diff
@@ -0,0 +1,59 @@
...
+    -arch=$(SM_ARCH) \
+    -O2 \
+    -Xcompiler -Wall \
+    -Xcompiler -Wextra \
+    -Xcompiler -Wno-unused-parameter \
+    -lineinfo \
+    $(INCLUDES)
+
+PRODUCER_CU  := ../../monitoring/csrc/ring/producer.cu
+DRAIN_SRC    := ../../monitoring/csrc/ring/drain_thread.cpp
```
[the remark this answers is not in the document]
zaoxing: Fixed in 726e45c. All three Makefile paths and the four `#include` lines in `test_rings.cu` now point at `native/csrc` — seven references in total; the `.cu` includes had the same problem and were not in the original comment.

Deliberately a rename and nothing else: normalise the two path spellings and all fourteen changed lines collapse to seven identical pairs, so no CUDA code, build flags, or logic are touched. Every referenced file exists under `native/csrc/ring/` with an unchanged name.

Verified by resolving each Makefile path and each `-I`-relative include. **Not compile-verified** — no nvcc on the dev machine, so this needs a CUDA box before anyone trusts that `make` succeeds.

Same root cause as the two stale Python test modules removed earlier in this branch: paths that survived the #95 src-layout move and returned with the Phase 4 commit. Worth noting separately that `test_null_mode.cu` includes `ring/pinned_pool.h` and `ring/chunk_assembler.h`, which exist nowhere in the repo — it is also not a Makefile target, so a path fix cannot help it.

Samfisheryu:
PR #113 exposes an interface-boundary issue before the new capture storage can be wired into the GPU path.

The CUDA ring/drain is mostly storage-agnostic, but the p2p → host boundary is hard-coded to `DMXHostEngine` and `ClickHouseRow`. We should refactor it to:

`p2p → backend-neutral CapturedTensor → CaptureSink`

Then keep the current behavior through `ClickHouseSink`, and connect #113 through `PackSink`. @XbzOnGit

tests/ring/Makefile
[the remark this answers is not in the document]
zaoxing: Update: this is now moot. 3ceb813 removed the entire `tests/ring/` suite from this branch — Makefile and all four `.cu` files — so the paths my earlier reply repointed no longer exist. The branch is now CPU-only: no `.cu`, `.cuh`, or `native/` file is added or modified anywhere in it.

Leaving the original finding recorded because it was correct, and because the same root cause (paths that survived the #95 src-layout move) produced two other regressions on this branch that were fixed rather than deleted.

zaoxing:
Agree with the direction.

One thing worth flagging: the coupling goes a bit deeper than the sink class. `dmx_host_queue_item` holds a `ClickHouseRow`, and `bindings.cpp` already flattens model_id / req_id / act_name / layer_no / shard_rank into row cells before the handoff. So most of the work is moving serialization to after `CaptureSink`, with `CapturedTensor` being raw bytes plus metadata.

Two things I would nail down early:

- `CapturedTensor` needs the full descriptor. `CaptureMetadata` wants 21 fields, the row today carries about five. Whatever is missing, `PackSink` cannot rebuild.
- Ownership at the seam. `PackSink` appends into reusable slabs, so a borrowed view with a clear lifetime keeps us from adding a memcpy on the hot path.

Also, if we plan to run both sinks side by side to compare: `index_version` comes from the indexing process own clock, so two writers can go non-monotone and a watermark can drop one side rows. It is in the Phase 5 limitations. Probably worth deciding before we start the comparison rather than during.

FWIW `PackSink` already has something to validate against. `tests/tools/golden_workload.py` emits a manifest over a corpus covering every dtype, and `tests/_faults.py` covers the failure behaviour.

src/dmi/storage/capture/clickhouse_catalog.py
```diff
@@ -0,0 +1,228 @@
...
+model_revision String, adapter_revision Nullable(String),
+capture_policy_version String, hook_name LowCardinality(String), layer_number Int32,
+producer_rank UInt32, step_number UInt64, token_start UInt64, token_end UInt64,
+batch_position UInt32, dtype LowCardinality(String), shape Array(UInt32),
+captured_at_ns UInt64, pack_id UUID, store_id LowCardinality(String), object_key String,
+object_bytes UInt64, pack_checksum FixedString(64), pack_record_count UInt32,
+payload_offset UInt64, stored_length UInt64, decoded_length UInt64,
+codec LowCardinality(String), payload_checksum FixedString(8), index_version UInt64,
+{_facet_ddl()}
+) ENGINE = ReplacingMergeTree(index_version)
```
Samfisheryu, line 127: **Blocker: pinned snapshots disappear after a normal ClickHouse merge.**

This table is used as version history by the watermark reader, but `ReplacingMergeTree(index_version)` eventually keeps only the highest version for each sorting key and removes the older rows.

I reproduced this on ClickHouse 25.12 with two identical copies of one descriptor:

1. Insert the descriptor at versions 1 and 2.
2. `get_by_ids(..., watermark="1")` returns one row.
3. Run `OPTIMIZE TABLE ... FINAL` to force the normal merge behavior.
4. The same watermark-1 lookup returns zero rows.

Observed result:

- raw rows: `2 → 1`
- watermark-1 result: `1 → 0`

`OPTIMIZE` only makes the reproduction deterministic; background merges can do the same automatically. This means a saved cursor or selection can stop resolving at an unpredictable time.

If old watermarks must remain readable, the history needs to be append-only, or the snapshot design must not depend on versions that this engine deletes.

src/dmi/storage/capture/clickhouse_reader.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def current_watermark(self) -> str:
+        """The catalog's high-water ``index_version``.
+
+        ``index_version`` is not part of the sort key, so this is a full-column
+        aggregate rather than an index lookup -- cheap in a columnar engine, but
+        not free, and it runs once per search.
+        """
+        rows = self._client.execute(
+            f"SELECT max(index_version) FROM {self._qualified()}",
```
Samfisheryu: **Blocker: this watermark becomes visible before the logical index batch is committed.**

`CatalogIndexer` assigns one `index_version`, then writes descriptors through multiple INSERTs of up to 10,000 rows before writing the pack markers. Reading `max(index_version)` from the raw descriptor table exposes that version after the first INSERT, while the rest of the same batch is still being written.

Reproduction:

1. Write 2 of 4 descriptors at version 123.
2. Read page 1 with `limit=1`; it returns watermark 123 while the table contains 2 rows.
3. Write the remaining 2 descriptors at version 123.
4. Continue using the pinned cursor.

Observed result:

- rows visible when the snapshot started: `2`
- rows returned by the pinned walk: `4`

Each ClickHouse INSERT may be atomic, but this Python loop is not one transaction. The reader can therefore observe and pin a half-written batch.

We need a separate published watermark: write every descriptor batch and pack marker first, then publish version 123 as the final step. Readers should use only that published version.

src/dmi/storage/capture/clickhouse_catalog.py
zaoxing: Confirmed, reproduced on 26.9.1:

```
BEFORE merge: raw rows = 2
  watermark 1 -> [(capture-a, 64)]
AFTER  merge: raw rows = 1
  watermark 1 -> []
```

You are right that this is a design problem rather than a bug. `argMax(...) WHERE index_version <= W` needs rows that `ReplacingMergeTree` is defined to delete, so the snapshot has an expiry nobody controls.

My live test missed it because it wrote two versions and queried immediately, never forcing a merge. It confirmed the mechanism I built instead of testing what the engine does to the rows underneath it. My fault.

One angle for the fix: descriptors come from immutable pack footers, so a re-indexed row is byte-identical to the one it replaces. The version history has no content difference in it. The watermark only really needs to exclude captures indexed after the snapshot, which may not require keeping superseded versions at all. Will come back with options rather than patch it blind.

src/dmi/storage/capture/clickhouse_reader.py
zaoxing: Confirmed, reproduced against the real indexer path with four descriptors written as two INSERTs at one version:

```
rows visible when the snapshot was taken : 2
rows returned by the pinned watermark    : 4
```

Agreed on the fix: write every descriptor batch and pack marker first, publish the version as a separate final step, and have readers use only the published value.

This was unreachable in my tests because `max_rows_per_insert` defaults to 10,000 and every corpus I used was small enough to fit one INSERT, so a batch never split. Adding coverage that forces multiple inserts per version.

src/dmi/storage/capture/clickhouse_reader.py
```diff
@@ -0,0 +1,386 @@
...
+            )
+        if not capture_ids:
+            return ()
+        params = {
+            "watermark": _parse_watermark(watermark),
+            "capture_ids": list(capture_ids),
+        }
+        sql = (
+            f"SELECT {self._projection()} FROM {self._qualified()} "
+            "WHERE capture_id IN %(capture_ids)s AND pack_id IN "
```
Samfisheryu: **Production scaling blocker: this point lookup cannot use the table ordering.**

The capture table is ordered by `(tenant_id, experiment_id, run_id, captured_at_ns, capture_id)`, but this query only knows `capture_id`. Because the ID is the last key, ClickHouse cannot jump directly to it.

I reproduced the lookup shape on ClickHouse 25.12 with 100,000 rows and the same ORDER BY. Looking up one ID produced:

```
PrimaryKey
  Keys: capture_id
  Parts: 1/1
  Granules: 12/12
```

So finding one capture scanned every granule. At one billion captures, hydrating ten selected IDs can still scan a large part of the billion-row catalog. This matters on every `estimate()` and `hydrate()`, because both resolve the selection through `get_by_ids`.

The lookup needs an access path ordered by identity, for example a projection or companion table on `(tenant_id, capture_id)`. The selection should carry the tenant as well.

src/dmi/storage/capture/catalog.py
```diff
@@ -0,0 +1,362 @@
...
+                raise ValueError(
+                    "catalog batch exceeds max_estimated_bytes: "
+                    f"{estimated_bytes + pack_bytes} > "
+                    f"{self._config.max_estimated_bytes}"
+                )
+            estimated_bytes += pack_bytes
+            descriptors.extend(pack_descriptors)
+            valid_refs.append(ref)
+
+        version = self._clock_ns()
```
Samfisheryu: **Production blocker for multiple indexers: version ordering is only local to one process.**

The new `_published_version` guard fixes repeated or backward clock values inside one `CatalogIndexer` instance, but a second process starts with no knowledge of the first one. Its wall clock can publish a lower version later.

I reproduced this against the updated implementation with two independent indexers:

1. Indexer A publishes two captures at version 200.
2. A reader pins watermark 200.
3. Indexer B runs later but publishes another capture at version 100.
4. The reader continues the original watermark-200 cursor.

Observed result:

```
rows when snapshot started       : 2
rows returned by the pinned walk: 3
```

The new row is version 100, so it falls underneath the already-pinned watermark 200 and appears inside the old snapshot.

This limitation is already mentioned in the design notes and in the PR discussion, but it is still a correctness blocker for multiple workers, failover, or a restarted worker after clock rollback. Versions need to come from one shared monotonic allocator or a database-owned catalog generation, not each process wall clock.

Samfisheryu (changes requested):
This must wait until the interface between ClickHouse and the GPU side has been refactored.

docs/capture-storage-design.md
```diff
@@ -0,0 +1,909 @@
...
+
+The Python implementation is a **reference implementation and conformance
+suite**, permanently -- not a candidate production writer. The reason is
+structural rather than performance-related: the ring transport reconstructs
+tensors on a native callback thread specifically to avoid touching Python or the
+GIL, and `DMXHostEngine` receives pre-assembled rows from that thread. There is
+no hot-path caller for a Python pack sink, and creating one would reintroduce
+per-tensor GIL contention that the ring exists to avoid.
+
+The pack-and-upload plane will therefore be written in C++ alongside the
```
XbzOnGit, line 919: **Blocker: this production-writer plan does not use the interface this PR was waiting for.** The merged branch now provides `ring::RecordSink`, owned `RecordEnvelope` objects, and sink injection through `RingEngine.create_record()`. Building another writer alongside `ClickHouseInsertStage` leaves capture packs disconnected from `RecordRuntime`. Before merge, please implement `CapturePackRecordSink : ring::RecordSink`, define `flush_and_wait()` at the spool/object-store durability boundary, expose an explicit backend choice while retaining legacy ClickHouse as the default, provide compatible offline reader configuration, and add a real `RecordRuntime -> pack sink -> catalog -> search/hydrate` example and bitwise E2E test.

src/dmi/storage/capture/clickhouse_reader.py
```diff
@@ -0,0 +1,386 @@
...
+    def _filters(
+        self, query: CaptureQuery, *, watermark: int, after: CursorKey | None
+    ) -> tuple[list[str], dict[str, object]]:
+        # The snapshot is the set of packs committed at or before the
+        # watermark, not a range of descriptor versions. Descriptor rows are
+        # byte identical across re-indexing, so which version survives a merge
+        # is irrelevant -- but a version *range* over them is not durable,
+        # because ReplacingMergeTree deletes superseded rows.
+        clauses = [
+            "pack_id IN (SELECT pack_id FROM "
```
XbzOnGit: **Blocker: snapshot membership drops `store_id`.** Pack identity is `(store_id, pack_id)`, but this predicate—and `get_by_ids()`—matches only `pack_id`. If store A commits UUID X at version 1 and store B commits UUID X at version 2, a version-1 snapshot admits B's row because X appears in the subquery. Match tuples in both places: `(store_id, pack_id) IN (SELECT store_id, pack_id ...)`, and add a same-UUID/two-store regression.

src/dmi/storage/capture/reader.py
```diff
@@ -0,0 +1,278 @@
...
+                    core=core,
+                    scalars=dict(scalars),
+                    artifacts=artifacts,
+                    failures=failures,
+                )
+            )
+        return tuple(summaries)
+
+    def _resolve(self, selection: CaptureSelection) -> tuple[CaptureDescriptor, ...]:
+        resolved = self._catalog.get_by_ids(
```
XbzOnGit, line 385: **Blocker: constructible selections bypass the published-watermark boundary.** `_resolve()` ignores `selection_id` and `filter_hash` and forwards any UInt64 watermark. A caller can construct a selection with `watermark=2**64-1` and resolve rows after commit-log insertion but before `publish_watermark()`, defeating publish-last. Validate the selection envelope and require an actually published watermark; if selections cross a trust boundary, sign them or store them server-side.

src/dmi/storage/capture/s3.py
```diff
@@ -0,0 +1,432 @@
...
+        verify_pack_source(pack)
+
+        metadata = {
+            "dmi-format": "dmi-pack-v1",
+            "dmi-pack-id": pack.pack_id,
+            "dmi-sha256": pack.checksum,
+            "dmi-record-count": str(pack.record_count),
+            "dmi-created-at-ns": str(pack.created_at_ns),
+        }
+        with pack.open() as source:
```
XbzOnGit, line 248: **Blocker: the upload can use a different stream than the one that was verified.** `verify_pack_source()` hashes the first `pack.open()`, then this line opens the source again for upload. A `PackSource` that returns different same-length bytes on its second open is accepted, and a spool uploader can then delete the only correct local copy. Upload the exact immutable stream that was verified, or require and verify a server-side content checksum before spool deletion.

src/dmi/storage/capture/s3.py
```diff
@@ -0,0 +1,432 @@
...
+            raise
+        if not isinstance(response, Mapping):
+            raise PackIntegrityError("S3 returned an invalid HeadObject response")
+        return response
+
+    def _existing_ref(
+        self, pack: PackSource, key: str, response: Mapping[str, object]
+    ) -> PackRef:
+        size, metadata = self._parse_head(response)
+        expected = {
```
XbzOnGit, line 501: **Blocker: retry accepts metadata that reconciliation rejects.** This existing-object check omits `dmi-format`. An object with correct bytes and every listed field but no format marker is accepted by `put()` and `stat()`; the uploader then deletes the ready file, while `inspect()` rejects the remote object. Reuse one strict metadata validator for retry, stat, and inspection—including format, pack ID, checksum, size, record count, and timestamp—before deleting the local copy.

src/dmi/storage/capture/spool.py
```diff
@@ -0,0 +1,518 @@
...
+                path.unlink(missing_ok=True)
+            for parent in {path.parent for path in stale}:
+                self._fsync_directory(parent)
+            paths = tuple(sorted(self.root.rglob("*.dmi-pack.ready")))
+            self._bytes = sum(path.lstat().st_size for path in paths)
+            self._entries = len(paths)
+            self._peak_bytes = max(self._peak_bytes, self._bytes)
+        entries = tuple(self._entry(path) for path in paths)
+        for entry in entries:
+            if self._checksum(entry.path) != entry.checksum:
```
XbzOnGit: **Blocker: one corrupt ready file blocks every healthy upload.** `recover()` raises on the first checksum mismatch before either uploader reaches per-pack error handling. With one corrupt and one healthy ready file, upload aborts and the healthy pack is never attempted; every retry repeats the failure. Validate entries independently, quarantine/report corrupt files, and continue uploading healthy entries.

src/dmi/storage/capture/pipeline.py
```diff
@@ -0,0 +1,611 @@
...
+        if self._writer is None:
+            writer = self._writer_with(record)
+            self._adopt(writer, record.metadata, scope=scope, now_ns=now_ns)
+        elif scope != self._scope:
+            writer = self._writer_with(record)
+            emitted.extend(self.flush(FlushReason.SESSION))
+            self._adopt(writer, record.metadata, scope=scope, now_ns=now_ns)
+        else:
+            try:
+                self._writer.append(record)
```
XbzOnGit, line 221: **High: maximum pack linger is not enforced under continuous traffic.** Expiry is checked only when `queue.get()` returns `None`; while the queue remains nonempty, this append path can keep an expired pack open indefinitely. Flush an expired open pack before appending each record, preferably inside `PackAssembler.append()` so direct callers get the same guarantee, and add a continuously-backlogged regression.

XbzOnGit (dismissed):
Requesting changes for seven additional merge blockers reproduced on the current merged head. These do not duplicate the four existing review threads. The legacy ClickHouse path passes the public regression matrix, but the new capture-pack path needs the integration, identity, durability, and liveness fixes below before merge.

docs/capture-storage-design.md
Samfisheryu: The #114 `RecordSink` boundary can support this with a `CapturePackRecordSink` adapter. However, #113’s pack pipeline is Python while the sink path is native C++, so a direct bridge introduces GIL, copy, and lifetime issues. The interface fits, but the production pack writer still needs a native implementation.

src/dmi/storage/capture/clickhouse_reader.py
zaoxing: Fixed in caa97d1 — the selection now carries the tenant, exactly as you suggested.

`CaptureSelection` gains `tenant_id` (derived in `create()` from the descriptors, single-tenant enforced, folded into the selection identity as v3), `CaptureCatalog.get_by_ids` takes it, and the reader puts `tenant_id = %(tenant_id)s` first in the WHERE clause so the primary index prunes to the tenant range. `ensure_schema` also adds an idempotent `bloom_filter(0.01)` skip index on `capture_id` to prune granules inside a large tenant.

Measured on a live server, 200k rows across 4 tenants, one 100-id selection:

```
before:  PrimaryKey parts 5/5, granules 25/25   200,004 rows read
after:   PrimaryKey parts 2/5, granules  7/25    52,660 rows read
```

Under a 150k `max_rows_to_read` cap the pre-fix query throws Code 158 (TOO_MANY_ROWS) and the post-fix one succeeds — the `estimate()`/`hydrate()` failure mode you predicted. Reads now scale with the tenant instead of the catalog.

A live regression test (`test_selection_resolve_prunes_to_the_tenant_range`) seeds two bulk tenants and asserts via `EXPLAIN indexes = 1` that the PrimaryKey index selects fewer granules than the table holds.

One behavior change worth flagging: an empty selection now raises, since there is no tenant to bind it to.

I did not add a projection or companion table — the tenant prefix plus the skip index gets the access path onto the sort key without a second copy of the data. Happy to revisit if you would rather have the projection.

src/dmi/storage/capture/catalog.py
zaoxing: Fixed in b86ac01 — versions are now allocated by the catalog, not by any process clock.

A new append-only `{prefix}_capture_version_claims` table (`version UInt64, claim_id UUID, claimed_at_ns UInt64`) backs `ClickHouseCatalogWriter.allocate_version()`, a sole-claimant protocol:

1. pick a candidate above `max(claims)` and the published watermark
2. insert a claim for it under a fresh UUID
3. read that version's claims back — proceed only if you are the sole claimant

Any contest is abandoned by everyone who sees it, then retried higher with a randomized skip (bounded by `allocation_attempts`). On a server with read-your-writes visibility the later of two same-version inserts always observes the earlier one, so at most one claimant ever sees a singleton: versions are unique and monotonic across processes, restarts, crashes and clock rollback. Every returned version is durably in the claims table, so later allocations always start above it. `claimed_at_ns` and `published_at_ns` are diagnostic only — no wall clock participates in ordering. Replicated setups would need `select_sequential_consistency`; that is noted in the docstring.

`CatalogIndexer` validates the allocation and refuses a non-monotonic one rather than silently bumping.

Tests in `tests/test_capture_version_allocation.py` cover concurrent allocation uniqueness, contested-version abandonment in both directions, attempt exhaustion, restart, and clock rollback across two indexers. The important one is live: `test_a_pinned_watermark_is_isolated_from_a_later_skewed_indexer` in `test_capture_end_to_end_live.py` runs indexer B with a deliberately lower clock after a reader pins W, and asserts the pinned read excludes B's captures while a fresh watermark includes them.

Residual, and I would value your read on it: an indexer still mid-batch at version V when a later W is published can land commit rows under a snapshot already pinned at W. Closing that means bounding membership by published versions rather than committed ones — a reader-side semantic change I deliberately did not make here.

src/dmi/storage/capture/clickhouse_reader.py
zaoxing: Fixed in 4624c10. Both query sites now match the composite identity against the commit log:

```sql
(store_id, pack_id) IN (SELECT store_id, pack_id FROM ... WHERE index_version <= %(watermark)s)
```

so a second store's copy of the same pack UUID can no longer satisfy membership for a snapshot that never committed it. SQL-shape tests pin the composite predicate on both `search()` and `get_by_ids()`.

The related half of your finding — that non-byte-identical descriptor rows for one logical capture would let a ReplacingMergeTree merge delete the row a pinned snapshot needs — is not closed by this change. Nothing violates the one-pack-per-capture invariant today, but nothing enforces it either, so that stays open on the sort-key thread.

src/dmi/storage/capture/reader.py
zaoxing: Fixed in 4624c10. `get_by_ids()` now rejects a selection whose `catalog_watermark` exceeds `current_watermark()`, mirroring the bound `decode_cursor` already enforced on the paginated path, so a hand-constructed selection can no longer read committed-but-unpublished rows. Covered by a unit test that constructs a selection above the published watermark and asserts the ValueError.

src/dmi/storage/capture/spool.py
zaoxing: Fixed in 89ebc11. `recover()` now validates each ready file independently: a corrupt one is renamed `*.quarantined` (bytes kept for offline inspection) and a file removed by a concurrent uploader is treated as completed work, so neither aborts the pass. One bad file no longer blocks the healthy backlog forever.

That also closed the race you would have hit next: the checksum loop runs outside the spool lock, so an overlapping uploader's `remove()` used to surface as a raw `FileNotFoundError` that failed the whole batch.

`test_spool_recovery_quarantines_corrupt_ready_pack` asserts quarantine-and-continue rather than the old fail-all behavior.

src/dmi/storage/capture/s3.py
zaoxing: Fixed in ce28b57 — the upload stream is now the verified stream.

`put()` opens the source exactly once and wraps it in a `_HashingReader` tee that computes SHA-256 over the bytes the transfer manager actually reads (`seekable()` returns False on purpose, which forces sequential reads so the tee sees every byte once and in order). After the upload, the observed byte count and digest must equal the pack declaration or the object is deleted (best effort) and `PackIntegrityError` is raised — so the spool keeps the local copy rather than discarding it after a bad upload. The separate `verify_pack_source()` pre-read is gone.

The contract is now stated at the call site: `PackRef.sha256 == SHA256(bytes actually uploaded)`, enforced by hashing the upload stream itself; the `dmi-sha256` object metadata is a stored assertion for later identity checks, not proof of server-side verification.

The regression test is deliberately adversarial: a `PackSource` whose `open()` returns different bytes on each call. It asserts the source is opened exactly once and that the bytes the client stored hash to the declaration — verifying A and uploading B is now structurally impossible rather than merely unlikely.

src/dmi/storage/capture/s3.py
zaoxing: Fixed in fe9e16f. The existing-object check on retry now requires `dmi-format == "dmi-pack-v1"` alongside the other markers, matching what `inspect()` demands, so a retry can no longer accept an object that reconciliation would reject forever — and, more to the point, can no longer let the spool delete the local copy of a pack that will never index. Regression test included.

src/dmi/storage/capture/pipeline.py
zaoxing: Fixed in 4624c10. `PackAssembler.append()` now calls `flush_expired` before appending, which is the placement you suggested, so `max_linger_ns` holds even when the queue never goes idle — and direct callers of the assembler get the same guarantee, not just the pipeline drain loop. The regression test drives a steady trickle of records that never reach the size or record caps and asserts the pack is flushed at the linger bound.

docs/capture-storage-design.md
zaoxing: Agreed with both of you, and I have not tried to close this in this PR.

Samfisheryu is right that the interface fits but the bridge does not: #113's pack pipeline is Python and the `RecordSink` path is native C++, so a direct adapter brings GIL, copy and lifetime problems that belong in a native implementation rather than a docs edit here. The design doc's production-writer section is the thing that is wrong today — it describes a Python writer as the production path when the merged #114 boundary says otherwise.

My read on sequencing: the storage format, catalog, and read path in this PR are independent of which side of the boundary produces the packs, and a `CapturePackRecordSink` written in C++ can target the same immutable pack format unchanged. So I would rather land this as the storage substrate and do the native writer as its own PR against the #114 interface, with the doc updated to describe the Python pipeline as the offline/ingest path rather than the production one.

If you would both prefer the doc corrected before this merges, I am happy to make that edit here — it is a docs change, not a code one. What I do not want to do is hand-wave a Python-to-native bridge into this PR.

src/dmi/storage/capture/s3.py
zaoxing: Follow-up: the checksum question is settled with live data, and I have added server-side verification in 742f781.

I brought up a real Garage 2.3.0 locally and probed it:

- `ChecksumSHA256` is supported **and enforced** — a deliberately wrong digest is rejected with `InvalidDigest`.
- It works on the single-part path, the multipart path, and with the non-seekable stream `_HashingReader` presents (that was the compatibility risk).
- For a multipart object the stored value is S3's composite checksum-of-checksums, not the whole-object digest — measured, they differ; single-part uploads do match. So the server checksum is deliberately never compared against `PackRef.checksum`.

The important part is that the two mechanisms catch different failures, so this is defence in depth rather than a replacement:

- the client-side tee catches a source whose bytes contradict its own declaration (the shapeshifting-source case);
- `ChecksumAlgorithm=SHA256` catches corruption between the bytes boto read and the bytes the server stored, part by part.

I confirmed the split experimentally: an upload whose stream corrupts itself mid-flight is still accepted by the server, because boto computes the part digests over the same corrupt bytes it sends. Only the tee catches that one — which is why it stays.

Also fixed while there: the live harness pinned `cargo:2.3.0`, a form only source builds report, so it rejected exactly the released binary the docs tell you to download. It now accepts either.

The Garage live contract passes against the released 2.3.0 binary, and a loopback sweep (4 packs x 32 MiB, 4 workers, 16 MiB multipart) sustained ~0.32-0.40 GiB/s with zero retries.

src/dmi/storage/capture/reader.py
```diff
@@ -0,0 +1,281 @@
...
+                    raise PackIntegrityError(
+                        f"object store returned a short range: "
+                        f"{len(block)} != {read_range.length}"
+                    )
+                for index in read_range.descriptor_indexes:
+                    descriptor = descriptors[index]
+                    start = descriptor.locator.offset - read_range.offset
+                    end = start + descriptor.locator.stored_length
+                    payload = block[start:end]
+                    verify_payload(descriptor, payload)
```
Samfisheryu, line 248: **High: catalog metadata is not bound to the pack footer.** Real ClickHouse repro: a `float32 (2,)` pack containing `[1, -2]` was re-described with the same locator/CRC as `int32 (1, 2)`. Hydration and decoding succeeded as `[[1065353216, -1073741824]]`, with unchanged payload bytes. Fetch/cache the authoritative footer and compare complete metadata plus locator fields before hydration (or verify a footer-anchored record digest); enforce the same check in `PackReader`.

src/dmi/storage/capture/s3.py
```diff
@@ -0,0 +1,505 @@
...
+        return self._existing_ref(pack, key, uploaded)
+
+    def _delete_uploaded(self, key: str) -> None:
+        # Best effort: the object contradicts its own metadata, so inspect()
+        # would reject it anyway; removal is cleanup, not correctness -- and a
+        # raised delete error must not mask the integrity error.
+        try:
+            self._client.delete_object(Bucket=self._bucket, Key=key)
+        except Exception:
+            pass
```
Samfisheryu, line 331: **High: failed cleanup lets a retry bless a corrupt S3 object.** With a wrong same-length upload and `delete_object()` returning 500, the first `put()` raises, but a correct-source retry returns `PackRef`; `stat()` claims the expected checksum while bytes differ and `PackIndex` reports `invalid trailer`. `inspect()` trusts the same metadata. Verify the actual existing-object digest, or conditionally publish the canonical key from a validated temporary object; never trust it after cleanup fails.

src/dmi/storage/capture/spool.py
```diff
@@ -0,0 +1,600 @@
...
+            except (PackIntegrityError, ValueError):
+                if path.exists():
+                    self._quarantine(path)
+                continue
+            if checksum != entry.checksum:
+                self._quarantine(path)
+                continue
+            entries.append(entry)
+        with self._lock:
+            self._bytes = sum(entry.object_bytes for entry in entries)
```
Samfisheryu: **High: `recover()` overwrites concurrent `stage()` accounting with a stale snapshot.** I blocked checksum validation after recovery listed pack A, staged pack B, then let recovery finish. Accounting fell from two packs back to one; a third pack was accepted with `max_bytes=1832`, leaving three ready files totaling `2748` bytes while `snapshot().bytes` still reported `1832`. This defeats the spool bound and can fill the disk. Hold the lock through a consistent recovery pass, or track a mutation generation and retry or merge concurrent deltas instead of assigning this stale aggregate.

src/dmi/storage/capture/filesystem.py
```diff
@@ -0,0 +1,276 @@
...
+                delete=False,
+            ) as handle:
+                temp_path = Path(handle.name)
+                copy_pack_source(pack, handle)
+                handle.flush()
+                os.fsync(handle.fileno())
+            try:
+                os.link(temp_path, path)
+            except FileExistsError:
+                return self._existing_ref(pack, object_key, path)
```
Samfisheryu: **High: losing the link race acknowledges durability before any directory fsync.** In a deterministic repro, another writer created the identical destination and `os.link()` raised `FileExistsError`; `put()` returned a valid `PackRef` with zero calls to `fsync_path_to_root()`. If the winner dies after linking but before its own fsync, this caller has acknowledged a directory entry that power loss may erase. After validating the existing object, the loser must execute the same directory-chain fsync before returning. Also ensure a newly created store root is made durable by syncing its parent; the current helper stops at `root` itself.

src/dmi/storage/capture/spool.py
zaoxing: Confirmed and fixed in f7f1c42. You were right that the stale aggregate defeats the bound.

Every accounting mutation (stage, remove, quarantine, recovery commit) now bumps a generation counter under the spool lock. recover() snapshots the generation with its listing, validates checksums unlocked as before, and commits the aggregate only if the generation is unchanged — otherwise it re-runs the pass, bounded at 8 attempts, then falls back to one final pass holding the lock. Optimistic-first because holding the lock across full-file checksumming of the backlog would block staging behind it.

Your repro is now a regression test verbatim: checksum validation event-gated mid-pass, a second pack staged from another thread, recovery released — the committed accounting includes both packs and an over-budget third stage is rejected. Pre-fix it failed exactly as you described (accounting fell to the stale aggregate and the third pack was accepted). A second test drives constant churn to prove the bounded retry and the locked fallback, including quarantine under the held lock.

src/dmi/storage/capture/filesystem.py
zaoxing: Confirmed and fixed in f7f1c42, both halves.

Both existing-object branches — the early path.exists() hit and the os.link FileExistsError race — now run the same fsync_path_to_root chain after _existing_ref validates the object, before returning, so the loser never acknowledges a dirent the winner has not yet made durable. Re-fsyncing an already-durable chain is cheap and idempotent.

Your second point about the root is also taken: FilesystemPackStore and DurablePackSpool construction now call a new fsync_new_root() — the root itself fsyncs unconditionally; its entry in the parent is fsynced best-effort, since the parent is often a directory this process does not own (a system temp dir, a mount point) and refusing service on EACCES there would be worse than the residual risk. That trade-off is documented at the helper.

Regression tests pin all three put() paths (normal, exists, link-race) calling the chain fsync, and both constructors syncing their new roots.

src/dmi/storage/capture/s3.py
zaoxing: Confirmed and fixed in ad0b568 — and your repro also exposed a wrong comment: _delete_uploaded claimed "the object contradicts its own metadata, so inspect() would reject it", which is false. The metadata was uploaded with the corrupt bytes and asserts the DECLARED checksum, so it agrees with the declaration; only the bytes lie. Corrected.

put()'s existing-object path now proves content rather than trusting metadata: the metadata comparison stays as a cheap pre-filter, then the object is streamed back in 1 MiB ranged reads and hashed against the pack declaration. A digest mismatch raises PackConflictError, so the staged file is kept and your scenario becomes an operator-visible upload failure instead of silently blessed corruption with the local copy deleted. The cost lands only on the idempotent-retry path, which is rare by construction; honest retries after a transient failure re-verify and succeed.

Your exact sequence is the regression test: same-length wrong bytes uploaded, delete_object failing with a 500, then a retry with the correct source — pre-fix it returned success; now it raises with the spool entry still present.

src/dmi/storage/capture/reader.py
zaoxing: Confirmed and fixed in 87a3426 — the catalog-driven hydration path was indeed the one place the footer was not consulted (PackReader.read_payload already refuses descriptors that contradict it).

hydrate() now loads each pack's footer once via PackIndex.from_store — the two small range reads, trailer and footer — before any payload read, and requires the catalog descriptor to equal the footer record in full: complete metadata plus offset, stored/decoded length, codec, and payload checksum. A mismatch raises PackFormatError naming the capture and the differing domain, before any payload bytes are fetched. Footers are cached per (store_id, pack_id, checksum) in a bounded LRU of 128 packs on the reader, so repeated hydrations of the same pack pay once — measured ~1 ms per uncached pack, nothing measurable on a hit. estimate() still reads nothing.

Your float32 (2,) -> int32 (1, 2) re-describe is a regression test at both levels: a unit test against the canned client, and a live ClickHouse variant that re-describes through a real second insert at a higher allocated version and published watermark — both raised nothing pre-fix and raise PackFormatError pre-payload now. The design doc's reader section now states the trust split explicitly: the catalog is the query index, the footer is the authority.

src/dmi/storage/capture/pack.py
```diff
@@ -0,0 +1,553 @@
...
+            indexed.checksum,
+        )
+        actual = (
+            descriptor.locator.offset,
+            descriptor.locator.stored_length,
+            descriptor.locator.decoded_length,
+            descriptor.locator.codec,
+            descriptor.locator.checksum,
+        )
+        if actual != expected:
```
Samfisheryu, line 513: **High: `PackReader.read_payload()` still accepts metadata that contradicts the footer.** Repro: the pack footer describes `float32 (2,)` bytes for `[1, -2]`; pass the same capture ID, locator, and CRC as `int32 (1, 2)`. `read_payload()` succeeds and `decode_tensor()` silently returns `[[1065353216, -1073741824]]`. The new `CaptureReader` check closes its path, but this method compares only locator fields. Please also require `descriptor.metadata == indexed.metadata`.

src/dmi/storage/capture/reader.py
```diff
@@ -0,0 +1,363 @@
...
+            ref = descriptors[plan.descriptor_indexes[0]].locator.pack_ref
+            # The catalog is the query index, but the pack footer is the
+            # authority for what each payload *is*: verify_payload checks
+            # length and CRC32 of the bytes only, so a re-described catalog
+            # row (same bytes, different dtype/shape) would otherwise decode
+            # garbage. Binding every catalog descriptor to the footer costs
+            # two small range reads (trailer + footer) per uncached pack per
+            # hydration -- the price of the footer being authoritative -- and
+            # happens before any payload range is fetched.
+            footer = self._footer_descriptors(store, ref)
```
Samfisheryu: **Blocker: cold footer verification bypasses the hydration limits.** With one 8-byte payload and `byte_limit=8, request_limit=1`, `hydrate()` succeeds, while the store records 3 requests and 796 bytes: trailer, footer, then payload. `estimate()` also reports only 1 request / 8 bytes. Footers may be 64 MiB, so this breaks the bounded-read contract. Include uncached footer requests and bytes in the estimate and both gates (and byte-bound the cache), or enforce a separate explicit metadata budget. The current gate tests warm the footer cache first, so they miss this cold path.

src/dmi/storage/capture/filesystem.py
```diff
@@ -0,0 +1,317 @@
...
+    promised durability for, while an openable parent still gets synced.
+    """
+
+    fd = os.open(root, os.O_RDONLY)
+    try:
+        os.fsync(fd)
+    finally:
+        os.close(fd)
+    try:
+        parent_fd = os.open(root.parent, os.O_RDONLY)
```
Samfisheryu: **High: recursively creating a store root is still not durably acknowledged.** `mkdir(parents=True)` can create several ancestors, but this helper fsyncs only `root` and its immediate parent, then silently accepts any parent-open error. Repro with `base/new-a/new-b/store`: only `store` and `new-b` were synced; `new-a` and the first stable parent `base` were not, so power loss can still remove the acknowledged store. Record the nearest existing ancestor before `mkdir`, then fsync every created level through that ancestor; a required durability sync should fail closed (or require a pre-created root), not be best-effort.

src/dmi/storage/capture/pack.py
Samfisheryu: Fixed in 32acc52. `PackReader.read_payload()` now requires the complete `CaptureMetadata` to match the authoritative footer before checking the locator. The float32 `(2,)` -> int32 `(1, 2)` repro now fails with `PackFormatError`; regression coverage is included.

src/dmi/storage/capture/reader.py
Samfisheryu: Fixed in 32acc52. The zero-I/O estimate now includes a cold-cache upper bound for both footer requests and bytes; hydration charges each trailer/footer read before issuing it, verifies every footer before fetching payloads, and bounds the LRU by count and serialized bytes. The repro is now bounded at 3 requests / 924 estimated bytes (3 / 796 actual), and an insufficient limit fails without a payload over-read.

src/dmi/storage/capture/filesystem.py
Samfisheryu: Fixed in 32acc52. Construction now fsyncs the complete path through a stable same-filesystem boundary on every call and fails closed on stat/open/fsync errors. Using a stable boundary also closes the failed-first-attempt retry and concurrent-constructor windows that a nearest-existing-ancestor check would miss. Both deterministic regressions are covered.

Samfisheryu (approved):
The adapter is now wired end-to-end as an explicit opt-in path. It is separate from the production ClickHouse sink, so the default production path remains unchanged and there is no double write. @XbzOnGit

The two catalog issues Alan identified still remain: out-of-order multi-indexer publication, and cross-pack duplicate captures under ReplacingMergeTree. These limitations should stay explicit and need further design discussion.

I’m approving this as a reference implementation. Whether to merge it now with these known limitations is Alan’s call.

zaoxing:
For the duplicate packs, we may need to outline the cases when this can happen. Does the CUDA writer will retry? Does this pack copy to another store? Does the pack migrate or repack?

zaoxing:
To solve both potential issues, I can see AI-patches probably won't work well. I propose we will do an overhaul like Apache Iceberg to address the potential duplicates with a log chain.

PR #114, opened 2026-08-26
[Core] Add framework-neutral record offload runtime

XbzOnGit:
## Summary

Add an explicit, opt-in record offload path for external DMI integrations. Integrations define their own hooks, record schemas, and descriptor encoding, while DMI core owns capture, transport, payload association, flushing, and schema-driven storage.

## What changed

- Add `HookPointV1` and ordered producer plans for integration-defined outputs.
- Add `RecordFormat`, `RecordSchema`, `RecordDescriptor`, and `RecordRuntime`.
- Extend the CUDA ring transport with tensor and scalar producers defined by those plans.
- Associate ordered payload tasks with integration-provided record descriptors.
- Support oversized CPU-direct fallback without changing the normal ring path.
- Add schema-driven ClickHouse tables and typed record insertion.
- Add checked, nonterminal durable flushing through the native host pipeline.
- Track reclaimable reservations per task; exact-size tasks bypass reclaim tracking.
- Export and document the additive surface through `dmi.api.v1`.

## Compatibility

The new path is selected only through explicit record-runtime and record-storage construction.

Existing behavior remains unchanged for:

- `MonitoringEngine` and the legacy ring setup
- fixed `HookPoint` definitions and presets
- the existing eight-cell inference record format
- Hugging Face integration
- the released vLLM integration

DMI core contains no Megatron-specific fields, schemas, selectors, or runtime policy.

## Validation

- Fixed CUDA 12.8 native builds: PASS
- Pre-rename focused Python regression: 72 passed, 6 deselected
- Final-head rename-focused Python regression: 44 passed
- Host-native regression: 49 passed
- CUDA native ring regression: 673 assertions passed, 0 failed
- Direct ClickHouse storage tests: 7 passed
- Generic ring-to-ClickHouse E2E: 1 passed
- Final-head existing HF/vLLM compatibility E2E: 7/7 cases passed

The compatibility matrix covered HF eager TP1/TP2 and vLLM eager TP1/PP1, TP2/PP1, TP1/PP2, plus vLLM CUDA Graph TP1/TP2.

## Scope

This PR adds only the framework-neutral core contract. Framework adapters, framework-owned schemas and metadata, orchestration, and model-specific hook placement remain in their integration repositories.

### DIFF
--- a/native/csrc/clickhouse_record_sink.cpp
+++ b/native/csrc/clickhouse_record_sink.cpp
@@ -59,7 +59,6 @@ std::vector<std::int64_t> resolve_shape(const ring::PayloadSlice& slice,
                                         std::uint64_t length_bytes,
                                         at::ScalarType dtype) {
     std::vector<std::int64_t> shape = slice.logical_shape;
-    if (shape.empty()) invalid("tensor payload requires a logical shape");
 
     const auto element_bytes =
         static_cast<std::uint64_t>(at::elementSize(dtype));
--- a/native/csrc/clickhouse_record_sink.cpp
+++ b/native/csrc/clickhouse_record_sink.cpp
@@ -1,0 +1,230 @@
+#include "clickhouse_record_sink.h"
+
+#include <ATen/ATen.h>
+
+#include <limits>
+#include <stdexcept>
+#include <utility>
+
+namespace dmx_host {
+
+namespace {
+
+[[noreturn]] void invalid(const std::string& message) {
+    throw std::runtime_error("clickhouse record sink: " + message);
+}
+
+std::uint64_t checked_product(const std::vector<std::int64_t>& shape,
+                              std::int32_t skip_dimension) {
+    std::uint64_t result = 1;
+    for (std::size_t i = 0; i < shape.size(); ++i) {
+        if (static_cast<std::int32_t>(i) == skip_dimension) continue;
+        if (shape[i] < 0) invalid("negative logical tensor dimension");
+        const auto dimension = static_cast<std::uint64_t>(shape[i]);
+        if (dimension != 0 &&
+            result > std::numeric_limits<std::uint64_t>::max() / dimension) {
+            invalid("logical tensor shape overflows uint64");
+        }
+        result *= dimension;
+    }
+    return result;
+}
+
+at::ScalarType checked_dtype(std::int32_t encoded) {
+    if (encoded < 0 ||
+        encoded >= static_cast<std::int32_t>(at::ScalarType::NumOptions)) {
+        invalid("unsupported encoded dtype");
+    }
+    const auto dtype = static_cast<at::ScalarType>(encoded);
+    if (at::elementSize(dtype) <= 0) {
+        invalid("dtype has no materializable element size");
+    }
+    return dtype;
+}
+
+std::uint64_t resolve_slice_length(const ring::PayloadSlice& slice,
+                                   std::uint64_t payload_bytes) {
+    if (slice.offset_bytes > payload_bytes) {
+        invalid("payload-slice offset exceeds actual payload bytes");
+    }
+    const std::uint64_t available = payload_bytes - slice.offset_bytes;
+    const std::uint64_t length = slice.length_bytes.value_or(available);
+    if (length > available) {
+        invalid("payload-slice length exceeds actual payload bytes");
+    }
+    return length;
+}
+
+std::vector<std::int64_t> resolve_shape(const ring::PayloadSlice& slice,
+                                        std::uint64_t length_bytes,
+                                        at::ScalarType dtype) {
+    std::vector<std::int64_t> shape = slice.logical_shape;
+    if (shape.empty()) invalid("tensor payload requires a logical shape");
+
+    const auto element_bytes =
+        static_cast<std::uint64_t>(at::elementSize(dtype));
+    if (length_bytes % element_bytes != 0) {
+        invalid("payload-slice bytes are not divisible by dtype size");
+    }
+    const std::uint64_t elements = length_bytes / element_bytes;
+
+    if (slice.inferred_dynamic_dim >= 0) {
+        if (slice.inferred_dynamic_dim >=
+            static_cast<std::int32_t>(shape.size())) {
+            invalid("inferred dynamic dimension is out of range");
+        }
+        const std::uint64_t fixed =
+            checked_product(shape, slice.inferred_dynamic_dim);
+        if (fixed == 0 || elements % fixed != 0) {
+            invalid("payload bytes cannot infer the requested tensor dimension");
+        }
+        const std::uint64_t inferred = elements / fixed;
+        if (inferred >
+            static_cast<std::uint64_t>(
+                std::numeric_limits<std::int64_t>::max())) {
+            invalid("inferred tensor dimension exceeds int64");
+        }
+        shape[static_cast<std::size_t>(slice.inferred_dynamic_dim)] =
+            static_cast<std::int64_t>(inferred);
+    } else if (checked_product(shape, -1) != elements) {
+        invalid("fixed tensor shape does not match payload-slice bytes");
+    }
+    return shape;
+}
+
+at::Tensor materialize_tensor(const at::Tensor& payload,
+                              const ring::PayloadSlice& slice,
+                              std::uint64_t length_bytes,
+                              at::ScalarType dtype) {
+    const std::vector<std::int64_t> shape =
+        resolve_shape(slice, length_bytes, dtype);
+    const auto element_bytes =
+        static_cast<std::uint64_t>(at::elementSize(dtype));
+    if (slice.offset_bytes % element_bytes != 0) {
+        invalid("payload-slice offset is not aligned to its dtype size");
+    }
+    at::Tensor bytes = payload.narrow(
+        0, static_cast<std::int64_t>(slice.offset_bytes),
+        static_cast<std::int64_t>(length_bytes));
+
+    // The ClickHouse host queue may outlive this envelope.  Other sinks can
+    // consume the owned payload directly without paying this per-cell copy.
+    return bytes.view(dtype).reshape(shape).clone();
+}
+
+}  // namespace
+
+ClickHouseRecordSink::ClickHouseRecordSink(SubmitRowFn submit_row,
+                                           FlushFn flush,
+                                           RethrowFn rethrow)
+    : submit_row_(std::move(submit_row)),
+      flush_(std::move(flush)),
+      rethrow_(std::move(rethrow)) {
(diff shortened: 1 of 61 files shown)


native/csrc/ring/record_consumer.cpp
```diff
@@ -0,0 +1,297 @@
...
+        materialize_and_submit(descriptor, payload);
+        {
+            std::lock_guard<std::mutex> lock(mu_);
+            --active_payloads_;
+        }
+        idle_cv_.notify_all();
+    } catch (...) {
+        {
+            std::lock_guard<std::mutex> lock(mu_);
+            --active_payloads_;
```
Samfisheryu, line 77: This is a durable-completion race. The catch path makes `active_payloads_` zero and wakes waiters before `failure_` is latched. `wait_until_idle()` can satisfy its idle predicate and `finish()` can return success while the submit callback is failing.

Repro on the current head: I used a `RecordConsumer` whose submit callback throws, with 64 concurrent durable-flush waiters calling `wait_until_idle()` followed by `finish()`. It reproduced immediately:

```text
false durable success at iteration 0
```

Please latch the first failure and decrement `active_payloads_` under the same `mu_` critical section, then notify. Please also add a regression test proving durable completion can never report success after a submit failure.

src/dmi/transport/ring.py
```diff
@@ -313,6 +367,120 @@ def submit_cpu_direct(self, cpu_tensor: torch.Tensor,
...
+            ends = output.producer_meta[1].detach().cpu().reshape(-1)
+            if starts.numel() == 0 or starts.numel() != ends.numel():
+                raise ValueError("SEGMENTED_PACK start/end length mismatch")
+            pieces = []
+            rows = source.size(0)
+            for start, end in zip(starts.tolist(), ends.tolist()):
+                begin = max(0, min(rows, int(start)))
+                finish = max(begin, min(rows, int(end)))
+                if finish > begin:
+                    pieces.append(source[begin:finish, ...].contiguous())
```
Samfisheryu: The CPU-direct fallback is not byte-equivalent to the GPU producer. This branch slices complete dim-0 rows and ignores `entry.transport_args[0]` (`feature_bytes`), while the CUDA kernel treats `feature_bytes` as the flattened row width.

Single-GPU repro on the current head:

```python
source = torch.arange(8, dtype=torch.float32).reshape(4, 2)
starts = torch.tensor([0], device="cuda")
ends = torch.tensor([2], device="cuda")
feature_bytes = 4
```

Observed persisted payloads:

```text
normal GPU path:      [0.0, 1.0]            8 bytes
oversized CPU path:   [0.0, 1.0, 2.0, 3.0] 16 bytes
```

Changing ring capacity can therefore change stored data. Please either validate `feature_bytes` against the physical tensor geometry or make the CPU fallback follow the exact CUDA byte semantics. The parity test should cover both `SEGMENTED_PACK` and `SEQ_PREFIX_PACK`.

src/dmi/records.py
```diff
@@ -0,0 +1,538 @@
...
+
+        For an oversized entry the transport has already flushed older work;
+        this method publishes the descriptor and completes the matching
+        CPU-direct submission.  The hook therefore skips its producer launch.
+        """
+
+        self._validate_entry_output(entry, output)
+        descriptor = self._encode(metadata, entry)
+        reservation = StepReservation(
+            self._transport.reserve_record(self._reservation_items((entry,)))
```
Samfisheryu, line 324: Generic record emission does not honor the engine capture-disabled state. `set_capture_enabled(False)` makes every CUDA record producer a no-op, but this path still reserves ring entries and publishes descriptors. No matching physical task can arrive, so durable flush waits until timeout.

Single-GPU repro on the current head:

```python
engine.set_capture_enabled(False)
hook(torch.arange(4, dtype=torch.float32, device="cuda"))
engine.flush_and_wait(0.25)
```

Observed:

```text
TimeoutError: timed out waiting for record ring completion
```

Please short-circuit record reservation/descriptor publication consistently when capture is disabled, including the replay path, and add a lifecycle regression test for disable -> hook/replay -> flush -> re-enable.

Samfisheryu (commented):
The HookPointV1 design looks good to me.

Samfisheryu:
@XbzOnGit I added the backend-neutral RecordSink boundary and moved the existing ClickHouse path behind an adapter in b1dda65. It is based on your latest correctness fixes. Could you take another look?

native/csrc/ring/producer.cu
```diff
@@ -318,6 +320,303 @@ __global__ void producer_chunked_kernel(
...
+        uint32_t sample = 0;
+        while (sample + 1 < batch &&
+               row >= static_cast<uint64_t>(
+                   valid_prefix_sum_dev_ptr[sample + 1])) {
+            ++sample;
+        }
+        const uint64_t sample_row = row - static_cast<uint64_t>(
+            valid_prefix_sum_dev_ptr[sample]);
+        const uint64_t source_row = sample_row * batch + sample;
+        record_copy_row_block(
```
Samfisheryu, line 585: `TransportSpec.feature_bytes` accepts any positive byte width, but this path reaches `d2d_copy_grid_stride`, which always uses 16-byte `uint4` copies. Rows after the first need not be 16-byte aligned.

Single-GPU repro with valid inputs: `[S=2, B=2, feature_bytes=20]`, `valid_count=[2,2]`, `valid_prefix_sum=[0,2,4]`. `torch.cuda.synchronize()` consistently reports `CUDA error: misaligned address`. `CHUNKED` has the same result with two 33-byte chunks selecting `[17,17]`.

Please keep the vectorized path only when both addresses are 16-byte aligned and use a byte-copy fallback otherwise. Add regression cases for 20-byte rows and odd chunk boundaries.

native/csrc/clickhouse_record_sink.cpp
```diff
@@ -0,0 +1,230 @@
...
+        invalid("payload-slice length exceeds actual payload bytes");
+    }
+    return length;
+}
+
+std::vector<std::int64_t> resolve_shape(const ring::PayloadSlice& slice,
+                                        std::uint64_t length_bytes,
+                                        at::ScalarType dtype) {
+    std::vector<std::int64_t> shape = slice.logical_shape;
+    if (shape.empty()) invalid("tensor payload requires a logical shape");
```
Samfisheryu: An empty shape is also the valid shape of a 0-D PyTorch tensor; it does not mean the shape is missing. The public runtime accepts `torch.tensor(3.5, device="cuda")` and records `output_shape=()`, but the end-to-end flush fails here with:

```text
RuntimeError: clickhouse record sink: tensor payload requires a logical shape
```

Please allow `shape == {}` and materialize it as a scalar tensor, with an end-to-end regression test.

src/dmi/hooks/record.py
```diff
@@ -0,0 +1,352 @@
...
+            outputs = tuple(inputs)
+        else:
+            outputs = _normalize_outputs(
+                spec.preprocess(*inputs), len(spec.outputs)
+            )
+        for index, (output_spec, output_id, value) in enumerate(
+            zip(spec.outputs, self._output_ids, outputs)
+        ):
+            output = _as_hook_output(value)
+            reservation = runtime.prepare_output(
```
Samfisheryu, line 266: `prepare_output()` may reserve ring space and publish the descriptor before `_dispatch()` validates the required producer metadata. For example, a valid `PREFIX_STRIP` spec with a `HookOutput` missing `row_count` first raises `ValueError: record producer requires row_count tensor`; the following flush then fails with `record flush found incomplete producer reclaims`. One input error therefore leaves this runtime unusable.

Please validate the required metadata count, dtype, device, dimensionality, and compatibility before `prepare_output()` performs reservation/publication. Add a regression asserting validation failure publishes neither a reservation nor a descriptor.

Samfisheryu (commented):
Three functional blockers found on the final head; details inline.

XbzOnGit:
All six functional review issues have been addressed on the latest head (`e74abda`):

- The record consumer now latches submit failure and decrements `active_payloads_` under the same mutex before notifying waiters. Concurrent durable-completion regression coverage was added.
- Oversized CPU-direct packing now follows the same flattened `feature_bytes` semantics as the GPU producer for both `SEQ_PREFIX_PACK` and `SEGMENTED_PACK`.
- Capture-disabled record emission now skips reservation and descriptor publication in both eager and replay paths, with disable/re-enable lifecycle coverage.
- Generic record producers now retain the vectorized copy only for aligned source and destination addresses and use a byte-copy path otherwise. Coverage includes 20-byte rows and odd chunk boundaries.
- Empty logical shape is now accepted as a valid 0-D tensor shape, with native and ClickHouse E2E scalar coverage.
- Producer metadata validation runs before reservation/publication in debugging mode, supports IDENTITY hooks, and is omitted from the release hot path by design. The duplicated Python producer chunk limit was also removed so the native constant remains the single source of truth.

Validation completed successfully:

- Focused record-hook tests: 17 passed.
- Generic record ClickHouse E2E: PASS.
- Full HF/vLLM ClickHouse-backed regression: 7/7 PASS across eager and CUDA Graph, TP1/TP2, and PP1/PP2.
- Final CUDA Graph TP2 comparison: 76,960/76,960 rows passed.
- PR CI `cpu-and-package-layout`: PASS.

native/csrc/ring/producer.cu
XbzOnGit: Fixed in e74abda. Generic record copies use the vectorized path only when both source and destination addresses are 16-byte aligned; otherwise they use a byte-copy path. CUDA regression coverage includes 20-byte packed rows and odd chunk boundaries.

native/csrc/clickhouse_record_sink.cpp
XbzOnGit: Fixed in e74abda. An empty logical shape is now treated as the valid shape of a 0-D tensor and materialized with `reshape({})`. Native sink coverage and ClickHouse E2E coverage both record and validate a CUDA scalar tensor.

src/dmi/hooks/record.py
XbzOnGit: Implemented in e74abda according to the agreed integration contract. In debugging mode, `_validate_producer_metadata()` runs before `prepare_output()`, validates metadata count, dtype, device,
  dimensionality, and compatibility, and supports IDENTITY hooks. The regression verifies that validation failure reaches neither reservation nor descriptor publication.

This validation is intentionally debugging-only: malformed producer metadata is a programmer/integration error that must be eliminated before release, and the release hot path does not pay Python validation overhead.

Samfisheryu (approved):
Re-reviewed the latest head. The previous functional blockers are fixed, and the focused Python, native CUDA, and ClickHouse E2E regressions pass. No remaining blockers.

PR #116, opened 2026-08-28
[Feat] Integrate Megatron-LM core_v0.17.1

XbzOnGit:
## Summary

Add Megatron-LM support to DMI through the standalone `DMI-Megatron-Integration` repository. DMI pins the integration repository as a submodule at commit `7838d16`, and that repository pins the matching `Megatron-LM-DMI` fork at commit `93f45d9`, based on Megatron-LM `core_v0.17.1`.

The integration repository owns the Megatron adapter, hook definitions, metadata propagation, runtime activation, record schemas, ClickHouse materialization and reconstruction, and Megatron-specific validation. DMI core remains framework-neutral.

## What changed

- Add `third_party/DMI-Megatron-Integration` as a reproducibly pinned Git submodule.
- Connect DMI to the Megatron integration package and its matching Megatron-LM fork without copying framework-specific code into DMI core.
- Restore the selective-recomputation gate contract and its documentation and tests: rejected occurrences publish nothing, while accepted zero-row records still publish their descriptor-backed zero-byte tasks.

## Compatibility

The gate change affects only rejected generic record-producer occurrences. Existing accepted records, zero-row records, dynamic-size reservation reclamation, legacy inference producers, record schemas, and storage formats remain unchanged.

Existing Hugging Face and vLLM integrations continue to use their established paths. Megatron-specific behavior remains in `DMI-Megatron-Integration` and `Megatron-LM-DMI`.

## Validation

- Clean environment: Python 3.12.14, PyTorch 2.13.0+cu129, CUDA toolkit 12.9.1, and Transformer Engine 2.14.1.
- Complete Megatron-to-DMI-to-ClickHouse regression: 249 passed, four expected native-Megatron PP=2 CUDA Graph skips, and 0 failed.
- All five Megatron EP reconstruction E2Es passed through Megatron, DMI, ClickHouse write, database query, and reconstruction validation.
- Native record-producer regression: 93 passed.
- Native ring-engine regression: 79 passed.
- Native null-mode regression: 48 passed.
- Record runtime, format, zero-row, and actual-byte reclamation regressions: 17 passed.
- Megatron host replay-selection regressions: 2 passed.
- Megatron TP1 local-CUDA-Graph selective-recomputation DMI-to-ClickHouse E2E: 1 passed.
- Existing Hugging Face and vLLM DMI-to-ClickHouse compatibility matrix: 7/7 cases passed.

## Scope

This PR adds the Megatron integration entry point to DMI and includes the core gate correction required by that integration.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "third_party/vllm-integration"]
 	path = third_party/vllm-integration
 	url = https://github.com/ProjectDMX/DMI-vLLM-Integration.git
+[submodule "third_party/DMI-Megatron-Integration"]
+	path = third_party/DMI-Megatron-Integration
+	url = https://github.com/ProjectDMX/DMI-Megatron-Integration.git
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@
 > **hallucination analysis**, **distillation**, **activation steering**, and beyond. If you're interested, please [contact us](mailto:ynn1999@umd.edu,sixianx@umd.edu,zaoxing@umd.edu).
 
 > **Project Status — research preview.** DMI currently supports HuggingFace
-> and vLLM backends for Qwen3 / Llama3.1 and GPT-2-family. SGLang support is on the way. APIs may change. Contributions, bug
-> reports, and feature requests are welcome.
+> and vLLM inference for Qwen3 / Llama3.1 and GPT-2-family, plus Megatron-LM
+> training. SGLang support is on the way. APIs may change.
+> Contributions, bug reports, and feature requests are welcome.
 
 > **👀Technical Report Available:** https://arxiv.org/abs/2605.11093
 
@@ -27,8 +28,6 @@ stages of the model lifecycle.
 
 - **More backend support and models** — Bring DMI to **SGLang** and expand support for
   more widely used model families, including multimodal models.
-- **From inference to training** — Extend DMI's low-overhead inspection to
-  large-scale training framework. (**Megatron**).
 - **From observation to action** — Low-latency streaming/pluggable APIs enables more downstream applications like online monitoring, activation steering,
     distillation, and speculative decoding.
 - **Broader PCIe-aware scheduling** — Extend DMI's serving-first drain governor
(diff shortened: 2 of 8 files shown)


XbzOnGit:
Updated this PR with the Megatron-LM installation and usage documentation:

- Added Megatron-LM to the root README and installation guide alongside the HF and vLLM backends.
- Added `docs/megatron.md` covering installation order, DMI native rebuild conditions, training launch, hook selection, and persistence modes.
- Updated `DMI-Megatron-Integration` to `812ea8b`, whose installation guide records the required compatibility constraints, tested dependency configuration, pinned Transformer Engine source, and correct DMI build order.
- The integration's `core_v0.17.1-dmi` and `main` branches both point to the same integration documentation commit.

This follow-up changes documentation and the integration submodule revision only; it does not change runtime code.

third_party/DMI-Megatron-Integration
```diff
@@ -0,0 +1 @@
+Subproject commit 812ea8b240526a3216576d16965d9fb2a325d2fa
```
Samfisheryu: [Blocker] `router-topk` crashes when Megatron token dropping leaves a token with fewer than `topk` surviving routes.

The failing code is in the pinned fork: [`router.py:262-274`](https://github.com/ProjectDMX/Megatron-LM-DMI/blob/93f45d90561608ec61820498cd1159ee348108e2/megatron/core/transformer/moe/router.py#L262-L274). False entries are replaced with `num_experts`; after capacity dropping, `torch.topk` can return that sentinel and the following `torch.gather` uses it as an out-of-range index.

I reproduced this on one RTX 4090 using the existing tiny real-training/file-sink E2E command with:

```text
CUDA_VISIBLE_DEVICES=0 CUDA_LAUNCH_BLOCKING=1
--dmi-hook-selection router-topk
--moe-router-topk 2
--moe-token-dispatcher-type alltoall
--moe-expert-capacity-factor 0.25
```

The first forward pass fails at `router.py:274` with `CUDA error: device-side assert triggered`. The same one-GPU E2E without token dropping passes.

Please either represent dropped slots with safe gather indices plus zero weights, or reject this combination during startup, and add a regression test where fewer than `topk` routes survive.

XbzOnGit: Fixed and propagated through all three repositories:

- Megatron-LM-DMI: `626c5ddfc`
- DMI-Megatron-Integration (`core_v0.17.1-dmi` and `main`): `b0e141f`
- This PR: `14584c4`

The `router-topk` preprocess now replaces dropped-route sentinels with safe indices only for `torch.gather`, while preserving the sentinel expert IDs and forcing dropped-slot weights to zero.

Added a regression test covering tokens with fewer than `topk` surviving routes, including zero surviving routes.

I also reran the reported one-GPU RTX 4090 file-sink E2E with:

```text
CUDA_VISIBLE_DEVICES=0 CUDA_LAUNCH_BLOCKING=1
--dmi-hook-selection router-topk
--moe-router-topk 2
--moe-token-dispatcher-type alltoall
--moe-expert-capacity-factor 0.25
```

The training iteration completed successfully. Both MoE layers recorded 24 dropped slots, and every dropped-slot weight was zero. PR CI also passes.

Samfisheryu (approved):
Re-tested the reported router-topk capacity-dropping case on one GPU. Training now passes, dropped-route sentinel IDs are preserved, and all dropped-slot weights are zero. No remaining blockers.

PR #117, opened 2026-08-30
[Fix] Publish ring payload readiness with system release

### LINKED ISSUE #99: [Core Refactor] Task-ring publication lacks a valid GPU-to-CPU synchronization edge
## Summary

The native task ring does not establish a valid CUDA happens-before relationship when publishing GPU-written entries to the CPU.

Its publication sequence is:

```text
write payload and task metadata
-> __threadfence()
-> volatile ready_seq store
```

The CPU polls `ready_seq` with an acquire load. However, `__threadfence()` has device scope and the volatile store is not a system-scope release. The CPU acquire therefore has no matching release operation. Observing `ready_seq` does not formally guarantee that the associated metadata and payload are ready; PCIe ordering cannot replace the required CUDA memory-model edge.

## Why not publish every entry with release/acquire?

Replacing every publication with a GPU system-scope release and CPU acquire would provide the ordering, but a direct RTX 4090 prototype measured:

```text
current publication:       2.393 us
system release/acquire:    3.295 us
difference:               +0.902 us (+37.7%)
```

A single-counter `D2D + publish` benchmark found approximately:

| Payload | Overhead |
|---:|---:|
| 4-64 KiB | ~70% |
| 1 MiB | ~51% |
| 16 MiB | ~11% |
| 64 MiB | ~0.5% |

In Qwen3-0.6B vLLM workloads, per-entry system publication increased complete generation time by approximately 0.6-0.8% with `hidden-states` (28 publications per forward) and 6.7-9.8% with `vllm-full` (341 publications per forward).

## Proposed direction

All currently supported hooks have CPU-known payload sizes and execution order. For this path, keep metadata in a CPU FIFO and use:

- the existing GPU payload ring;
- a GPU-local produced-entry counter; and
- one CPU/GPU-shared monotonic `published_head`.

Publish `published_head` with a system-scope release every fixed `K` hooks and at every forward boundary. The CPU observes it with an acquire load and only then submits the corresponding D2H to the drain stream.

This replaces per-entry managed publication and polling with one release per batch and one polled cache line. It does not reduce payload D2D or D2H traffic, and it delays CPU visibility until a batch boundary. Publishing at every forward boundary ensures that a partial batch cannot remain hidden.

The required ordering chain is:

```text
Stream-A payload writes and writer join
-> GPU system release
-> CPU acquire
-> CPU submits Stream-B D2H
-> Stream-B reads the payload
```

The same CPU thread must acquire the published value before enqueueing the D2H; the mechanism cannot order a copy that was already enqueued. The counter must be a naturally aligned system-scope atomic in managed memory on a device with `concurrentManagedAccess == 1`. Publication must join every payload writer, and ring storage cannot be reused until its D2H completes.

### GPU-resolved payload sizes

Hooks whose identity and order are CPU-known but whose actual payload size is resolved on the GPU--for example, an EP per-expert output whose row count depends on GPU routing--can use an optional managed ring with one `uint64_t` size slot per outstanding dynamic entry. The CPU FIFO carries a reservation upper bound and the corresponding slot identity.

In eager execution, a hook that is dynamic only relative to pre-forward planning can instead enqueue the exact size just in time when it is available to the CPU at hook invocation. The managed size ring is only needed when the size remains GPU-resolved at publication time.

The GPU writes the actual size before the next batched release of `published_head`; the same release/acquire edge then publishes both payload and size, so the size slot needs no separate synchronization. Capacity uses the reservation upper bound, requires `actual_size <= reserved_size`, and prevents slot reuse until CPU consumption. Dynamic hook identity, count, and execution order remain out of scope.

## References

- [CUDA C++ memory model and system-scope atomicity](https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cuda-cpp-memory-model.html)
- [PTX release and acquire patterns](https://docs.nvidia.com/cuda/parallel-thread-execution/#release-and-acquire-patterns)
- [PTX memory and CUDA API synchronization](https://docs.nvidia.com/cuda/parallel-thread-execution/#memory-synchronization)

XbzOnGit:
Closes #99.

## Summary

Payload producers run on the framework CUDA stream, while the CPU drain thread enqueues D2H copies on an independent stream. The old protocol wrote several fields in CPU-preferred managed memory and then published a volatile ready marker after a device-scope fence.  
This left two ordering problems:
- CPU could observe readiness without a consistent metadata entry.
- Observation did not establish system-scope ordering between preceding GPU payload writes and the later cross-stream D2H copy.

This change replaces each metadata entry with one packed 64-bit publication word. Bit 63 is readiness and the remaining bits hold the actual tensor size. The actual size must still come from the GPU because dynamic producers, including EP hooks, may not know it on the CPU.

The last participating GPU block publishes the word with a system-scope release store after the existing writer-completion join. The CPU polls it with an acquire load and only then submits the D2H copy. The resulting chain is:

```text
GPU payload writes
-> writer-completion join
-> system release publication
-> CPU acquire observation
-> D2H submission on the drain stream
```

- Publishing the ready bit and actual tensor size in one atomic word prevents the CPU from observing readiness with inconsistent size metadata.
- The system-scope release/acquire handshake orders the GPU payload writes before the CPU submits D2H on the drain stream, establishing the required cross-stream ordering.

Offsets and wrap spans are reconstructed from CPU-authoritative ring state, so they no longer generate managed-memory publication traffic. No extra thread fence is needed in the publication helper: the release store itself orders the preceding payload work. Skipped hooks do not publish a slot, and public configuration, payload layout, schemas, and consumer output remain unchanged.

## Regression coverage

- Packed publication encoding, acquire polling, clearing, slot reuse, wraparound, dynamic-size producers, timed flushes, and multi-GPU worker binding.
- Independent vLLM reference capture comparing tensor inventory, dtype, shape, offsets, and payload bytes.
- Same-execution stress coverage to exclude differences from dynamic batching between separate runs.

## Validation

- Native ring suite: 508 distinct assertions passed with zero failures.
- Qwen3-0.6B, one GPU, 1 MiB rings: 54,566/54,566 checks passed; all 54,560 transported tensors were bitwise exact, and logprobs were bitwise exact for 8/8 prompts.
- Twenty sequential same-execution vLLM runs: 480,000/480,000 tensor comparisons passed.
- The final `sm_89` binary contains the expected system fence and system-scope publication store; the CPU acquire path was also inspected.

## Performance

Overhead relative to the corresponding no-DMI vLLM baseline:

| Hook selection | Before fix | After fix |
|---|---:|---:|
| Hidden states (`resid_pre`) | +5.03% | +1.99% |
| Complete `vllm-full` selection | +12.67% | +13.27% |

Setup: Qwen3-8B FP16, TP=2 on two RTX 4090 GPUs, vLLM CUDA graphs, Python 3.12.14, PyTorch 2.13.0+cu129, CUDA 12.9, and vLLM 0.27.1. Both implementations were clean-built with the same toolchain; initialization, graph capture, and terminal draining were outside the timed region.


native/csrc/ring/task_ring.cuh
```diff
@@ -56,93 +56,34 @@ inline uint64_t task_free_slots(
...
-    __threadfence();
-    *reinterpret_cast<volatile uint64_t*>(&entries[idx].ready_seq) =
-        READY_SEQ_SENTINEL;
+    const uint64_t idx = seq_no % capacity;
+    cuda::atomic_ref<uint64_t, cuda::thread_scope_system> ready(
+        slots[idx]);
+    // No separate __threadfence(): the system-scope release store orders all
+    // preceding payload writes before the CPU can acquire this publication.
+    ready.store(PUBLICATION_READY | actual_bytes,
+                cuda::memory_order_release);
```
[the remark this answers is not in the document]
XbzOnGit: This is intentionally encoded at the device publication site because encode_publication() is currently a host-side shared C++ helper; calling it here would require adding CUDA execution annotations. The packed-size constraint is unchanged because the helper performs the same OR without additional validation. We’ll keep the direct expression to avoid expanding the CUDA-facing interface.

zaoxing: **Cross-block payload visibility bug: the removed `__threadfence()` is still needed (or the arrival counter must become acq_rel).**

The comment here isn't quite right: a release store only orders writes that *happen-before it in this thread's view*. The publishing block's own payload writes are covered, but other blocks' payload stores reach the publishing block only through the last-block arrival counter — and that's a **relaxed** `atomicAdd` (producer.cu:125), which creates no happens-before edge on the reader side.

The old code was correct precisely because of the pairing that this PR deletes:

- each block: payload stores → `__threadfence()` → `atomicAdd(counter)` (producer.cu:184)
- last block: reads counter → `__threadfence()` (deleted here) → publish `ready_seq`

The fence-fence synchronization is what made every block's payload writes visible before publication. With it gone, for any launch with `n_blocks > 1` (any tensor above TIER0), block A's `uint4` payload stores may not yet be system-visible when block B release-stores `READY | actual_bytes`. `DrainThread::scan_ready` acquires the word and D2H-copies the payload region → silently corrupted tensor bytes delivered to the record sink.

Fix options:
1. Make the arrival counter a `cuda::atomic_ref<..., thread_scope_device>` `fetch_add` with `memory_order_acq_rel` (the acquire on the last block's read then brings all fenced writes into its view, and the release store's cumulativity publishes them), or
2. Reinstate a fence (`cuda::atomic_thread_fence(cuda::memory_order_acquire, cuda::thread_scope_device)` or `__threadfence()`) between the counter read and the release store.

Either way, the now-inaccurate comments at producer.cu:128 and producer.cu:137-139 should be updated to describe the actual synchronization chain.

---

native/csrc/ring/drain_thread.cpp
```diff
@@ -497,16 +510,17 @@ void DrainThread::scan_ready() {
...
     while (true) {
         if (pending_entries_ >= task_cap) break;
-        if (!task_cpu_ready(ring_.task_entries, task_cap, visible_head_)) break;
-
         const uint64_t idx = visible_head_ % task_cap;
-        TaskEntry ec = ring_.task_entries[idx];
+        uint64_t publication = 0;
+        if (!publication_try_acquire(
+                &ring_.publication_slots[idx], publication)) break;
+        const uint64_t actual_bytes = publication_size(publication);
```
zaoxing, line 517: **Dropping the per-slot sequence check turns a task-ring overrun from a detectable stall into silent payload corruption.**

The old `task_cpu_ready` accepted a slot only when its `ready_seq` equaled the exact sequence the drain expected, so a slot republished a lap early (capacity-check race, a second producer stream, or producer/drain counter desync after an error) read as *not ready* — the pipeline stalled visibly, and the sequence match in `account_record_task` stayed sound.

`publication_try_acquire` now accepts **any** word with `PUBLICATION_READY` set and attributes it to `visible_head_`. Under an overrun, that delivers the wrong `actual_bytes` under the wrong sequence, desynchronizing `pending_bytes_` / `cpu_payload_tail_` and mis-slicing the staging bytes of every subsequent task, with no error raised.

The 64-bit word has room to keep this defense at the same cost: `actual_bytes` needs nowhere near 63 bits, so a few high bits below `PUBLICATION_READY` could carry `seq_no % 2^k` (producer writes `READY | (seq << SIZE_BITS) | actual`, consumer checks the tag against `visible_head_` and treats a mismatch as a hard error or not-ready). Even a 1-bit lap parity would catch the common one-lap overrun.

---

zaoxing (commented):
Deep review of the publication-word rework. The span-reconstruction equivalence, scan/flush/clear ordering (`pending_entries_ >= task_cap` guard), zero-byte publication handling, and the `TaskEntry` removal all check out. Two correctness concerns worth addressing before merge, posted inline:

1. **task_ring.cuh — cross-block visibility**: the deleted `__threadfence()` was the acquire half of the fence/fence pairing with each block's pre-`atomicAdd` fence; the system-scope release store alone doesn't publish *other* blocks' payload writes, so multi-block producers can publish before the payload is visible (silent corruption).
2. **drain_thread.cpp — overrun detection lost**: accepting any READY word (vs. the old exact `ready_seq` match) converts a ring overrun from a visible stall into silent mis-sliced payloads; a few sequence-tag bits in the publication word would restore it.

Minor (no inline comment): the publication-word decode helpers are now hand-rolled in four places (drain_thread.cpp anonymous namespace, test_rings.cu, test_producer.cu, and the inlined encode in task_publish) — worth consolidating the CPU-side acquire/size/clear helpers into publication_word.h next to `encode_publication`, especially if sequence bits get added per point 2.

---

native/csrc/ring/task_ring.cuh
XbzOnGit: Fixed in f022947. The last block now executes a device-scope acquire fence after the completion atomicAdd and before the system-scope release publication. This pairs with every block’s pre-arrival __threadfence(), establishing:

payload writes → block fence → completion atomicAdd → last-block acquire fence → system release publication

The related comments were updated accordingly.

native/csrc/ring/publication_word.h
```diff
@@ -0,0 +1,16 @@
...
+#include <cstdint>
+
+namespace ring {
+
+static constexpr uint64_t PUBLICATION_READY = uint64_t{1} << 63;
+static constexpr uint64_t PUBLICATION_SIZE_MASK = PUBLICATION_READY - 1;
+
+constexpr uint64_t encode_publication(uint64_t actual_bytes) {
+    return PUBLICATION_READY | actual_bytes;
+}
```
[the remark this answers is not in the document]
XbzOnGit: Masking would not change the encoded value: READY | actual_bytes is identical to READY | (actual_bytes & SIZE_MASK) because READY already forces bit 63. If size validation is needed, it should reject an out-of-range value rather than silently truncate it. In practice, actual_bytes is bounded by the ring capacity and cannot approach 2^63.

native/csrc/ring/drain_thread.cpp
XbzOnGit: Thanks—the old `ready_seq` check was useful defensive detection: if a physical slot contained a later ring generation, the CPU treated it as not ready instead of silently consuming the wrong size and payload span. It detected the mismatch as a stall, but did not recover or propagate an explicit error.

The current protocol relies on five invariants:

1. CPU outstanding reservations never exceed task-ring capacity.
2. The CPU reclaims capacity only after acquiring and clearing the corresponding publication slot.
3. Every CPU reservation maps to one and only one GPU publication.
4. CPU accounting and GPU publications preserve reservation order, with producers executing on one CUDA stream.
5. CPU and GPU start from matching logical sequence state, advance once per task, and use the same `sequence % task_capacity` slot mapping.

Under these assumptions, two live tasks cannot reach the same slot without exceeding capacity. Assumptions 1, 2, and 5 are simple localized ring accounting/indexing; assumption 4 is stable once CPU accounting order matches the single GPU stream. Assumption 3 is the most fragile because it crosses hook estimation, conditional/dynamic hooks, Python integration, and GPU execution.

This should become more stable when vLLM and Hugging Face use `HookPointV1`. Eager execution can reserve and account each output just in time immediately before producer dispatch, giving identical CPU/GPU order as long as the producer publishes exactly once. For CUDA graphs, hook order and count are learned during capture and replayed from a fixed recorded schedule; a changed schedule must trigger recapture.

There are two complementary detection options:

1. **Per-slot generation tag:** pack `READY | generation | actual_size`, with `generation = sequence / task_capacity`. A `1/16/47` layout detects the first wrong ring generation without extra managed-memory traffic and still supports tensors below 128 TiB. It adds per-publication encoding/checking and the generation tag eventually wraps.

2. **Coarse cumulative audit:** compare GPU-authoritative tensor/byte totals with CPU totals at a synchronized forward, iteration, or every-N-forwards boundary. Existing `task_head` and `actual_bytes_counter` can provide the GPU view. This directly audits assumption 3, the most fragile invariant, without changing the publication format and amortizes communication. Its limitations are delayed detection, the need for a common frontier, and possible compensating errors that preserve aggregate totals.

The two can be combined: generation tags check slot identity immediately, while cumulative counters audit end-to-end count and byte accounting.

I propose handling this in a separate PR because useful detection requires complete fail-stop behavior and error propagation, not only a tag or counters. On mismatch, the drain must stop consuming/clearing slots and submitting D2H, latch the first error, wake waiters, propagate native non-zero/error status through bindings as an exception, and make the integration disable DMI or fail explicitly. That PR should include fault injection for both detection and propagation.

native/csrc/ring/producer.cu
```diff
@@ -119,34 +119,28 @@ __device__ inline void publish_last_block_arrives(
     uint64_t       task_head,
     uint64_t       payload_head,
     uint64_t       alloc_bytes,
-    const TwoSpan& spans,
     uint64_t       actual_total)
 {
     if (threadIdx.x != 0) return;
     uint32_t finished = atomicAdd(&g_block_done_counter, 1);
```
Samfisheryu, line 146: **Blocker: the completion counter still does not formally join payload writes from non-thread-0 workers.** Each writer executes `__threadfence()` before `__syncthreads()`, but only thread 0 performs this relaxed `atomicAdd`. CUDA fence-fence synchronization requires the release fence to be sequenced-before the counter atomic on the same thread; the block barrier does not turn another worker’s fence into that edge. Therefore the last-block acquire does not formally import every worker’s payload writes before READY is published. This passed 3×4090 stress (6,000 × 40 MiB), but hardware observation does not close the missing happens-before edge.

The minimal fix is to add a device-scope release fence here, after the callers’ `__syncthreads()` and before this relaxed `atomicAdd`, while retaining the last-block acquire and system release publication:

```cpp
if (threadIdx.x != 0) return;
cuda::atomic_thread_fence(cuda::memory_order_release,
                          cuda::thread_scope_device);
uint32_t finished = atomicAdd(&g_block_done_counter, 1);
```

Alternatively, use a device-scope acquire/release RMW for the arrival join.

native/csrc/ring/task_ring.cuh
XbzOnGit: Correction to my earlier reply: f022947 alone was incomplete because the worker-thread fences were not sequenced before thread 0’s relaxed arrival RMW. 85fcdfa completes the join by adding a device-scope release fence in thread 0 after the block-wide barrier and before atomicAdd, while retaining the last block’s device-scope acquire fence before system-release publication. The counter reset is now atomic, and the required single-producer-stream assumption is documented.

native/csrc/ring/producer.cu
XbzOnGit: Fixed in 85fcdfa as suggested. After every caller’s block-wide __syncthreads(), thread 0 executes a device-scope release fence before the relaxed completion atomicAdd. The last block then executes the device-scope acquire fence before system-scope release publication. This joins all workers’ payload writes into the publishing thread. The shared counter reset was also changed to atomicExch.

XbzOnGit:
## Implementation updates
The current head is 8776f14. 85fcdfa completes the block-to-block publication join, a497117 fixes padded transport accounting and wrap offsets, and 8776f14 documents the resulting size/release/reclaim contract. The native suite passes 621 assertions and CI is green.

For the generation/overrun detection proposal, my proposal is to defer it to a follow-up with fail-stop handling and error propagation. See that conversation for details.

## Benchmark
I also reran the TP=2 performance benchmark at the current head (`8776f14`) because `85fcdfa` and `a497117` affect the critical path.

Setup: Qwen3-8B FP16, TP=2 on two RTX 4090 GPUs, vLLM CUDA graphs, 8 prompts × 64 output tokens, two fresh processes per mode, 2 warmups and 5 measured iterations. The ring remained below the flush threshold, so no D2H occurred inside the measured region.

| Mode | Mean latency | Throughput | Overhead |
|---|---:|---:|---:|
| Baseline | 0.731423 s | 700.01 tok/s | — |
| Hidden states (`resid_pre`) | 0.740762 s | 691.18 tok/s | **+1.28%** |

The previously reported packed-publication implementation measured +1.99% overhead against its corresponding baseline. Given normal run-to-run variation, the updated result shows no measurable performance regression from the subsequent correctness fixes.

Samfisheryu (approved):
Reviewed the latest fixes. No remaining blockers under DMI's supported single-producer-stream contract.

zaoxing (approved):
LGTM

PR #119, opened 2026-09-01
Capture catalog: pack identity in the descriptor key, fenced snapshot publication

zaoxing:
## What this branch does

Reworks the ClickHouse capture catalog around pack identity and safe snapshot publication:

- **Pack identity in the descriptor sort key** — descriptors are keyed by (index_version, store_id, pack_id), so two packs describing the same capture are distinct rows and supersession is resolved by one argMax over the full tuple.
- **Snapshot manifest + publish-id pairing** — membership is a (index_version, publish_id) pair between manifest rows and a conditional watermark write, so a lost publish leaves only inert rows that no reader can see.
- **Durable publisher lease with in-statement fencing** — publication requires an exclusive, recoverable lease; every writing statement carries the fence server-side, so a fenced-out publisher writes nothing visible.
- **Schema-version stamp with refusal, not migration** — a build refuses to run against a catalog it cannot read, with a full rebuild procedure in the error and docs.
- **Publish retry loop** in the indexer, **exact integer order statistics** in tensor summaries (CORE_SUMMARY_VERSION 2), a **bounded public capture view**, and regenerated golden manifest.

## Review and fixes

A 10-angle adversarial review produced 15 verified findings. Follow-up audits found additional edge cases; all are fixed on this branch, including:

- Descriptor rows are re-stamped at the re-allocated version on publish retry, so argMax supersession tracks publish order (was: a batch published last could lose to one published earlier).
- The schema guard counts effective membership, refusing a catalog whose watermark table was truncated (was: silently empty catalog reporting success).
- `release_publisher_lease` is a single fenced INSERT…SELECT (was: check-then-act race that could revoke a successor's live lease).
- `commit_packs` runs before `SnapshotPublishConflictError` propagates (was: visible publishes silently republished).
- The version barrier handles the empty-table NULL case; the pinned-watermark read in `get_by_ids` is sequentially consistent; retry exhaustion raises `SnapshotPublishExhaustedError` inside the error taxonomy; `publish_timeout_ns` must be whole seconds so `max_execution_time` can never truncate to unlimited; manifest/identity lists are chunked under `max_query_size`; a no-op batch performs no catalog writes; plus a writer-owned `drop_schema()` and a single shared membership predicate to end drop-list and predicate drift.
- Later fixes quarantine contested lease heads, bind pack tenants to object keys, apply configured quorum to consistent descriptor and protocol writes, require object-scoped schema visibility and a complete post-DDL catalog before stamping, and make runtime storage-backend selection explicit.

## Testing

- Current CPU suite: 1063 passed locally; 6 pre-existing macOS host failures remain (OpenMP shared memory and tests that assume Linux `/proc/self/fd`).
- Focused catalog, reader, review-finding, and runtime suites: 321 passed on the current head.
- Infrastructure-backed runs recorded on this branch: 59 live ClickHouse tests; 62 with Garage/S3 configured; and a dedicated two-replica quorum harness.


tests/test_clickhouse_schema_migration_live.py
```diff
@@ -0,0 +1,820 @@
...
+    module = importlib.util.module_from_spec(spec)
+    # `dataclass` resolves annotations through `sys.modules`, so the module has
+    # to be registered before it is executed, and removed after so nothing else
+    # in the session can import the predecessor by accident.
+    sys.modules[name] = module
+    try:
+        spec.loader.exec_module(module)
+    finally:
+        del sys.modules[name]
+    return module
```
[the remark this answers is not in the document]
zaoxing: Fixed in c8875a8: the git invocation now skips on OSError, and a missing spec/loader skips cleanly instead of raising.

src/dmi/storage/capture/catalog.py
```diff
@@ -253,13 +355,57 @@ def index(self, refs: Sequence[PackRef]) -> IndexResult:
...
         # a batch lands underneath a watermark a reader already pinned and that
         # snapshot grows after the fact. The version is a DB-allocated
         # generation owned by the catalog itself -- no wall-clock value
         # participates in the ordering, so clock skew between indexers (or an
         # NTP step on one of them) cannot reorder publications. The wall clock
-        # only stamps published_at_ns below.
+        # only stamps published_at_ns.
         version = self._writer.allocate_version()
         if type(version) is not int or version < 0:
             raise ValueError("allocate_version must return a non-negative integer")
```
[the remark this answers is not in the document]
zaoxing: Fixed in c8875a8: the check now requires a positive integer and the message says so (last_published_version() >= 0, so a 0 already breaks the strictly-above-the-head contract). The suppressed summary.py comment is also addressed there: empty integer tensors now keep integer order statistics, pinned by a new test.

zaoxing:
## Request changes — three blocking findings

GitHub does not permit the PR author to submit a formal Request changes review, so these blockers are posted as a PR comment. Reviewed at `c8875a8`.

1. **Critical — a fenced write can finish after lease takeover** (`src/dmi/storage/capture/clickhouse_catalog.py:1099`).

   `publish_snapshot()` renews once, then executes every manifest chunk before issuing the watermark statement. A later statement can therefore start with less than `publish_timeout_ns` remaining on the lease. If it starts just before expiry, its fence passes, but it may remain in flight until after a successor acquires the lease. The old publisher's lower-version watermark can then land after the successor's publication, making a snapshot pinned at the successor's version grow afterward.

   Make every fenced statement fail closed unless `expires_at_ns > now + publish_timeout_ns`, or renew immediately before every manifest chunk and the watermark statement. Add a regression covering a statement that begins near expiry and completes after takeover.

2. **Required — partial membership loss passes the schema guard** (`src/dmi/storage/capture/clickhouse_catalog.py:944`).

   `_inventory_without_membership()` only rejects when effective membership is completely empty. If one effective member survives while watermark or manifest rows for other inventory packs are lost, `members` remains nonzero and `ensure_schema()` accepts the catalog. `committed_pack_ids()` then continues skipping every inventory pack, so the missing packs stay permanently invisible.

   Check for any inventory identity lacking effective manifest membership, using an anti-join over `(store_id, pack_id)`, rather than comparing only global nonzero counts. Add coverage where some, but not all, inventory packs retain effective membership.

3. **Required — the 1,000-tuple chunks can still exceed `max_query_size`** (`src/dmi/storage/capture/clickhouse_catalog.py:267`).

   Chunking by tuple count does not establish the stated byte bound. `S3StoreConfig` permits a 255-byte `store_id`; with the installed `clickhouse-driver`, 1,000 valid `(store_id, pack_id)` members serialize to 301,000 bytes before the surrounding SQL, already above ClickHouse's 256 KiB default. Both manifest publication and `committed_pack_ids()` can therefore still fail with Code 62 for valid inputs.

   Chunk by the encoded statement-byte budget, or enforce a shared identity-size bound and derive a conservative tuple limit from it. Add a boundary test using maximum-length valid store IDs.

Verification at this head: 256 targeted catalog, lease, reader, and summary tests pass. The current suite does not exercise the timing, partial-membership, or maximum-encoded-size cases above.

XbzOnGit:
Please retain HTTPS URLs for these public submodules. `.gitmodules` defines the portable read path for all users and automation; public HTTPS clones are anonymous, while SSH requires GitHub key setup. The Actions jobs pass because `actions/checkout` rewrites these URLs to HTTPS, but ordinary recursive clones and other automation do not. Developers who prefer SSH can configure Git’s `url.*.insteadOf` locally.

src/dmi/storage/capture/clickhouse_lease.py
```diff
@@ -0,0 +1,267 @@
...
+        self._lease = None
+
+    def release_statement(self) -> str:
+        table = self._qualified_table
+        return (
+            f"INSERT INTO {table} "
+            "(term, lease_id, holder, acquired_at_ns, expires_at_ns) "
+            "SELECT term + 1, toUUID(%(lease_id)s), %(holder)s, "
+            "toUnixTimestamp64Nano(now64(9)), toUnixTimestamp64Nano(now64(9)) "
+            f"FROM (SELECT term, lease_id FROM {table} "
```
XbzOnGit: **Required:** this `INSERT ... SELECT` is not a compare-and-set relative to another concurrent insert. A stale release can read expired lease A at term T; claimant B can then insert T+1, complete its singleton read-back, and receive a live lease; finally A's already-running release can insert its expired row at the same T+1. That leaves a contested head until B's TTL, and when A's UUID sorts higher, B is fenced immediately.

I reproduced this on ClickHouse 25.12 by adding `AND sleep(1) = 0` to this release query, inserting a live successor after 200 ms, and using deterministic IDs. The resulting head was:

```text
(2, ffffffff-ffff-ffff-ffff-ffffffffffff, stale, expired)
(2, 00000000-0000-0000-0000-000000000001, successor, live)
resolved head: stale
```

The fake at `tests/_catalog_fakes.py:213` resolves and writes release together, so the current test cannot model this overlap. The simplest repair is to omit eager release and accept TTL expiry; otherwise this needs actual serialization or a release representation that cannot share/supersede the successor's term. Please add a live overlap regression.

src/dmi/storage/capture/pack.py
```diff
@@ -548,10 +549,78 @@ def _parse_records(decoded: dict[str, object], footer_offset: int) -> tuple[_Ind
...
+    for it, but the object-key pattern does not allow it. Left alone, an
+    identifier containing ``~`` produces a key every store rejects, and the
+    sink failure is fatal to the whole pipeline. An identifier too long for a
+    key segment is replaced by a digest of itself, which is one-way -- so the
+    comparison below has to be able to recognise that form too.
+    """
+    encoded = quote(value, safe="-_.=").replace("~", "%7E")
+    if len(encoded.encode()) <= 160:
+        return encoded
+    return "sha256-" + sha256(value.encode()).hexdigest()
```
XbzOnGit: **Required:** the direct and digest forms share a namespace, so this mapping is not one-to-one. Both IDs below satisfy the 512-byte metadata limit but produce the same object-key segment, which means `_reject_a_foreign_tenant()` cannot distinguish them:

```python
from hashlib import sha256
long_id = "t" * 300
literal_id = "sha256-" + sha256(long_id.encode()).hexdigest()
assert key_component(long_id) == key_component(literal_id)
```

Observed common value: `sha256-0afa5eb0871c38e72e8dfed71861ecde977c9635456a24bc289c0edeae50e1cc`. I also put a pack carrying `long_id` under the identical `tenant=<segment>` for `literal_id`; `PackIndex.from_store(...).descriptors()` accepted it. Please make the output forms disjoint—for example, tag every direct form as `raw-...` and digest forms as `sha256-...`—and pin this pair in a regression. Since that changes the v1 object-key layout, the backward-reading/migration behavior should be explicit.

tests/test_clickhouse_schema_migration_live.py
```diff
@@ -0,0 +1,1029 @@
...
+# first refusal read the absent version stamp as "this is version 1" and
+# recited version 1's two differences at it -- both false of that catalog. An
+# operator who checks two named facts, finds both untrue and concludes the
+# guard is broken goes around it, which is the one path the guard exists to
+# prevent. So the diagnosis is now read off the live schema, and this is the
+# test whose absence let that ship.
+
+# `@{upstream}` at the time of writing. Pinned as a sha rather than as the
+# symbolic ref, because `@{upstream}` is a per-clone alias that moves.
+_UPSTREAM_COMMIT = "bea3ed828f69de0a56b2bf30023a8a49fc19745e"
```
XbzOnGit, line 476: **Required:** this commit is reachable only from the current feature branch, not from `main`. `fetch-depth: 0` fixes shallow history but cannot retain an object after the PR is squash-merged and its branch is deleted. A full main-only clone reproduces it:

```bash
git clone --branch main --single-branch https://github.com/ProjectDMX/DMI.git /tmp/dmi-main
git -C /tmp/dmi-main cat-file -e bea3ed828f69de0a56b2bf30023a8a49fc19745e^{commit}
# fatal: Not a valid object name ... (exit 128)
```

This repository's PR #113 demonstrates the squash pattern: its source commits became single-parent `08ea071` on `main`. Once this fixture disappears, `_upstream_catalog_module()` skips and the workflow's following “Fail if any live test was skipped” step fails post-merge CI. Please check an immutable predecessor fixture into `tests/data` (optionally with a pinned checksum), or place it behind a permanent ref that CI explicitly fetches.

src/dmi/storage/capture/__init__.py
```diff
@@ -58,23 +58,35 @@
 from .summary import (
     CORE_SUMMARY_VERSION,
     ArtifactRef,
-    CoreTensorSummaryV1,
+    CoreTensorSummary,
```
XbzOnGit, line 61: **Required:** `CoreTensorSummaryV1` is exported from `dmi.storage.capture` on `main`, but this replacement removes that public import with no compatibility path while `pyproject.toml` remains at `1.2.0`:

```bash
PYTHONPATH=src python -c 'from dmi.storage.capture import CoreTensorSummaryV1'
# ImportError: cannot import name 'CoreTensorSummaryV1'
```

Bumping the serialized summary schema to 2 is reasonable, but it does not by itself preserve the Python API. Please retain/document a compatibility type or alias (with clear schema-version semantics), or make this an explicitly versioned breaking API change. An import-compatibility test would pin the intended decision.

docs/install.md
```diff
@@ -49,6 +49,20 @@ cd DMI
...
+every piece of automation inherits, and an HTTPS clone of these repositories
+needs no key setup. To use SSH instead, rewrite the remotes on your own machine
+rather than editing `.gitmodules` -- the rewrite applies to the superproject and
+every submodule, including nested ones, and survives a re-clone:
+
+```bash
+git config --global url."git@github.com:".insteadOf https://github.com/
+```
+
+If you already have a checkout and change either the URLs or that config, run
```
XbzOnGit: Minor documentation correction: `git submodule sync --recursive` is needed after `.gitmodules` URLs change, but not after changing only `url.*.insteadOf`. The rewrite is applied dynamically when Git resolves a stored URL. For example, an origin stored as HTTPS resolves immediately with `git -c 'url.git@github.com:.insteadOf=https://github.com/' remote get-url origin`, without a sync. Please separate the two cases so readers do not overwrite intentional local submodule URL settings unnecessarily.

XbzOnGit (changes requested):
Requesting changes for four reproducible correctness, CI-reliability, and API-compatibility issues. The ordinary validation is otherwise strong: I independently ran 1,063 CPU tests, 61 ClickHouse live tests, compile/package checks, and confirmed the current GitHub jobs are green. The required items and compact reproductions are inline; the `insteadOf` documentation note is minor.

src/dmi/storage/capture/clickhouse_lease.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def fence(self) -> str:
+        return (
+            "(SELECT count() FROM ("
+            "SELECT lease_id, expires_at_ns FROM "
+            f"{self._qualified_table} "
+            "ORDER BY term DESC, lease_id DESC LIMIT 1"
+            ") WHERE lease_id = toUUID(%(lease_id)s) "
+            "AND expires_at_ns > "
+            "toUInt64(toUnixTimestamp64Nano(now64(9))) "
```
XbzOnGit, line 208: **Blocking:** This fence compares an expiration timestamp produced by `now64()` on the host handling `_insert()` with a fresh `now64()` on the host handling the publish. In a multi-host replicated deployment, ClickHouse-host clock skew therefore affects lease ownership. The overlap condition is:

```text
clock_B - clock_A > lease_ttl_ns - publish_timeout_ns
```

For example, with a 2 s TTL, 1 s publish timeout, and host B 1.1 s ahead, B can consider A expired about 0.9 s after A renews while A may still be executing within its timeout. A fence evaluated before B's claim can then complete below B's later watermark. The replica verifier cannot exercise this because both replicas use one server process. Please use a shared lease/order authority, or enforce a measured host-skew bound and subtract it from the fence margin, with a real two-host offset test.

src/dmi/storage/capture/clickhouse_catalog.py
```diff
@@ -229,26 +185,360 @@ def write_descriptors(
...
+        # Manifest rows of a publish that never reached the watermark, at a
+        # version BELOW the published head. Below the head, its watermark
+        # INSERT can no longer land -- the barrier requires strictly above --
+        # so the pair the membership predicate needs will never exist. An
+        # in-flight publish always sits ABOVE the head, which is what keeps
+        # this from deleting membership out from under it.
+        removed[self._manifest] = self._delete_rows(
+            self._manifest,
+            "index_version < %(published)s AND (index_version, publish_id) "
+            f"NOT IN (SELECT index_version, publish_id FROM "
```
XbzOnGit: **Blocking:** This mutation reads an independently replicated table in its predicate. For `ReplicatedMergeTree`, ClickHouse does not permit this mutation form under the default settings, while `collect_garbage()` supplies only `mutations_sync=1`. On the tested 25.12 server, this check returned `0` for both settings:

```sql
SELECT name, value
FROM system.settings
WHERE name IN (
  'allow_nondeterministic_mutations',
  'mutations_execute_subqueries_on_initiator'
);
```

The Keeper-backed verifier never calls GC, so it does not cover this path. ClickHouse documents the relevant defaults and replicated-mutation behavior [here](https://github.com/ClickHouse/clickhouse-docs/blob/main/docs/managing-data/updating-data/overview.mdx#support-for-subqueries-and-non-deterministic-functions). Please resolve the orphan tuple set with a consistent initiating-side read and issue bounded literal-tuple deletions, or use another replica-consistent operation, then add GC coverage to the Keeper-backed verifier.

src/dmi/storage/capture/clickhouse_schema.py
```diff
@@ -0,0 +1,638 @@
...
+                "snapshot, so an indexing pass would skip them and report "
+                "success while their captures remain invisible. "
+                + self.rebuild_instruction()
+            )
+        return False
+
+    def _catalog_state(self) -> dict[str, CatalogObject]:
+        names = [name for _, name in self.objects + self.legacy_objects]
+        rows = self._client.execute(
+            "SELECT name, engine, sorting_key FROM system.tables "
```
XbzOnGit: **Blocking:** The compatibility state records only `engine` and `sorting_key`, so every non-view engine is treated as an acceptable table; it never checks the required `ReplacingMergeTree(index_version)` argument. I reproduced this on ClickHouse 25.12 by replacing a stamped v4 `capture_raw` with the same columns and key but `ReplacingMergeTree()`: `ensure_schema()` returned successfully.

I then inserted versions `3, 1` for one pack and version `2` for another, published membership only for versions 2 and 3, and ran `OPTIMIZE ... FINAL`:

```text
versions inserted: [3, 1, 2]
versions after FINAL: [(1,), (2,)]
reader-selected pack: version-3 pack before FINAL, version-2 pack after FINAL
```

Please validate the engine family and arguments through `engine_full` or normalized `create_table_query` before any DDL, and add a correct-key/wrong-engine live test.

src/dmi/storage/capture/clickhouse_schema.py
```diff
@@ -0,0 +1,638 @@
...
+        pack_view = self.qualified(self.pack_view)
+        watermark = self.qualified(self.watermark)
+        manifest = self.qualified(self.manifest)
+        self._client.execute(f"CREATE DATABASE IF NOT EXISTS {database}")
+        if fresh:
+            # Re-read what looked like nothing at all, now that the database
+            # exists: a second installer that created objects between the two
+            # reads is met with the same refusals as any other catalog rather
+            # than with this build's DDL running over its work.
+            self._verify_compatibility()
```
XbzOnGit, line 350: **Blocking:** This second empty-state read does not close initialization. Another initializer can create the old `capture_raw` after this read but before this initializer's `CREATE TABLE IF NOT EXISTS`:

```text
v4:  compatibility read -> empty
old: CREATE capture_raw with the pre-v4 sort key
v4:  CREATE TABLE IF NOT EXISTS -> no-op
v4:  confirm names/kinds and write schema stamp 4
```

I injected that interleaving on ClickHouse 25.12. The first `ensure_schema()` returned and wrote stamp 4 over the old sort key; only a second call detected and refused the state. The caller is therefore allowed to index into a catalog whose stamp and table layout disagree. Please serialize initialization for each catalog prefix. Re-running all compatibility checks immediately before stamping is useful additional protection, though it does not replace serialization.

src/dmi/storage/capture/clickhouse_catalog.py
```diff
@@ -229,26 +185,360 @@ def write_descriptors(
...
+        removed: dict[str, int] = {}
+        published = self.last_published_version()
+        head = self._leases.head()
+        settings = {"mutations_sync": 1}
+
+        # Manifest rows of a publish that never reached the watermark, at a
+        # version BELOW the published head. Below the head, its watermark
+        # INSERT can no longer land -- the barrier requires strictly above --
+        # so the pair the membership predicate needs will never exist. An
+        # in-flight publish always sits ABOVE the head, which is what keeps
```
XbzOnGit: **Major:** Please do not treat `index_version < published_head` as proof that a manifest can never become relevant. The protocol documentation already allows a lower watermark statement to pass its barrier and fence, remain in flight while a higher publisher completes, and then commit below the new head. If GC runs between those commits, it removes the lower manifest; the delayed watermark can still land and pass owner read-back, leaving a committed inventory entry with no readable snapshot membership.

I reproduced the resulting state on ClickHouse 25.12 with this ordering:

```text
insert manifest(low)
publish(high)
collect_garbage()
insert watermark(low)  # the already-admitted statement completes
commit_packs(low)
```

The lower manifest count changed from 1 to 0, the descriptor was absent at the published watermark, and the next `ensure_schema()` refused the inventory-without-membership state. Caveat: this models the ordering after the lower statement's guards have evaluated; it does not independently reproduce a blocked lower statement. Please retain an explicit finished/abandoned state, or eliminate late lower commits before GC may remove their manifests.

tests/tools/verify_replicated_quorum.py
```diff
@@ -0,0 +1,211 @@
...
+    #    insert is INVISIBLE to a select_sequential_consistency read (measured
+    #    on 25.12: plain read 2 rows, sequential read 1). Every read-back here
+    #    is a sequential read of the writer's own insert, so the protocol
+    #    cannot confirm itself and gives up rather than proceeding blind.
+    mixed = ClickHouseCatalogWriter(
+        client, ClickHouseCatalogConfig(database="default", table_prefix=kept_prefix)
+    )
+    try:
+        mixed.allocate_version()
+        print("FAIL  quorum turned off mid-life went unnoticed")
```
XbzOnGit: **Major:** This verifier can print `FAIL` and still exit successfully. An unexpected successful allocation only prints this message, while the negative branches accept any exception rather than checking the expected exception and ClickHouse code. The successful path never calls `write_descriptors()` and uses `refs=()`, so it exercises neither descriptor nor manifest writes; query-log evidence is optional, and GC is not tested.

Because the design document identifies this as the replicated protocol verifier, please make unexpected branches raise/assert, validate the expected server result, require the query-log evidence being tested, publish at least one descriptor reference, and exercise manifest GC.

docs/benchmarks.md
```diff
@@ -130,28 +130,40 @@ PYTHONPATH=src python -m benchmarks.bench_capture_search \
...
+resolved column ordered on `(index_version, store_id, pack_id)`. The A/B for
+that change alone, on this host: a 100-row page went 140.1 ms -> 171.7 ms,
+**+22.6%**, and +17% to +43% across page sizes, depth and selectivity. The
+per-column form left the winner of a tie undefined, so a pinned selection could
+resolve to a different pack after a background merge; the delta is what fixing
+that cost.
 
 The two snapshot rows are not alternatives. `FINAL` drops any capture
-re-indexed above the watermark, so the 1.85x gap is what correctness costs, not
+re-indexed above the watermark, so the 2.1x gap is what correctness costs, not
```
XbzOnGit: **Major:** The headline measurements do not run the production snapshot query. `measure_snapshot_shapes()` reads `max(index_version)` from `capture_raw` instead of `index_watermark`, uses separate `argMax` expressions instead of the production full-row ordering, omits manifest membership, and pins at the raw maximum so it never creates the historical-pin case described here.

Consequently, the reported 35.6 ms versus 16.8 ms comparison does not establish a 2.1x cost for the current reader, and the reported 2.3 ms "watermark" timing measures a descriptor-table aggregate. The page-latency measurements do call the real reader and are unaffected. Please benchmark the production query or reader entry point, use the real watermark and manifest, compare equal projections, and insert a descriptor above a historical pin.

XbzOnGit (dismissed):
Second-pass review of `10661d7`. I am requesting changes for four new Blocking findings and three new Major findings in the inline threads. This review does not repeat the earlier threads.

Validation completed:

- CPU suite: 1,063 passed, 1 skipped.
- Complete local ClickHouse 25.12 suite on port 9000: 61 passed.
- Focused migration suite: 12 passed.
- Targeted ClickHouse checks reproduced the schema-engine issue, the concurrent-initialization issue, and the resulting catalog state for the documented late-publish/GC ordering.
- Both current GitHub checks pass.

Minor follow-ups that do not independently drive this review state:

- **Minor:** `CatalogIndexer._reject_a_writer_that_cannot_publish()` trusts a cached non-`None` lease, so an old holder can allocate a version and write a batch before renewal notices that ownership changed.
- **Minor:** `consistent_snapshot_reads` does not require an actual Boolean; for example, the string `'false'` enables sequential consistency.
- **Minor:** The pinned `get_by_ids` test compares two identical post-takeover calls instead of comparing with a saved pre-takeover result.
- **Minor:** Several protocol-document statements still describe earlier behavior: exactly two publish statements, no manifest cleanup, and an undecided version source.

src/dmi/storage/capture/clickhouse_lease.py
zaoxing: Addressed in `607b474`, taking the second option you offered: the skew bound is now part of the fence margin.

`fence()` requires `publish_timeout_ns + clock_skew_ns` of lease life remaining. With the holder's host ahead of true time by `a`, a successor's by `b`, and `|b - a| <= S`: the holder is admitted only while more than `timeout + S` remains on its own clock and is capped at `timeout` on that same clock, so it finishes before true time `e - S - a`; the successor can claim only after `e - b`. The gap is `S - (b - a) >= 0`, so the two cannot overlap unless real skew exceeds the configured bound.

So that the bound cannot be silently absent where it matters, `CatalogConfig` now refuses `insert_quorum` with `clock_skew_ns == 0`, and requires `publish_timeout_ns + clock_skew_ns < lease_ttl_ns`.

**This does not close your finding.** The bound is *declared*, not *measured* — nothing checks the configured value against real host offset, so your overlap returns for an operator who under-declares. And the two-host offset test you asked for has not been run: every server this repository talks to is a single process, including the Keeper-backed verifier, whose two replicas share one clock. The CPU tests pin the SQL predicate, which is a statement about the SQL and not about hosts.

The procedure and the result that would confirm or refute it are written down in `docs/catalog-descriptor-key.md` under "Verification this repository cannot run" (item 1), so it is not being carried as done.

---

src/dmi/storage/capture/clickhouse_catalog.py
zaoxing: Fixed in `2940b4e`, the way you prescribed: a plain `SELECT` on the initiator resolves the orphan pairs, which are then deleted as literal tuples. No mutation reads a second replicated table any more.

The literal `IN` list is deterministic on every replica, and it is chunked under `max_query_size` so an unbounded orphan set cannot overflow the statement text. The code comment cites your finding directly — `allow_nondeterministic_mutations = 0` and `mutations_execute_subqueries_on_initiator = 0` on 25.12.

`2bda77f` adds `collect_garbage()` coverage to the Keeper-backed verifier, as you asked.

**Unverified.** The verifier has not been run since that rewrite — this environment has no Keeper. Recorded in `docs/catalog-descriptor-key.md` item 2, including the point that the retention step must be *admitted*: an `ALTER TABLE ... DELETE` that ClickHouse refuses would raise `ServerException` there. Running it is the check that actually closes this thread.

---

src/dmi/storage/capture/clickhouse_schema.py
zaoxing: Fixed in `0ea19d2`: installs are serialised per prefix on the publisher lease, which is the requirement you named. A writer that already holds the lease keeps it and renews around the install; otherwise the installer takes the lease itself and gives it back afterwards.

`36a6b09` adds the re-check immediately before the stamp — the additional protection you noted, not a substitute for the serialisation.

Two details worth flagging for your re-read:

- The two tables an installer needs before it can take the lease at all (the stamp table and the lease table) are created ahead of it. Both are idempotent and both are this build's own, so neither is part of the interleaving you injected.
- There is a deliberate fast path: when every object already exists, the install skips the lease. Nothing is created in that case, the DDL is idempotent over it and the stamp is a server-side no-op, so there is no interleaving for the lease to exclude. If you read that differently I would rather hear it now.

---

src/dmi/storage/capture/clickhouse_catalog.py
zaoxing: Addressed in `2940b4e`, though not by either of the two options you listed — worth saying plainly so you can judge whether it is enough.

Rather than an explicit finished/abandoned state, `publish_snapshot` now confirms its membership *after* the watermark row stands, and refuses the publish when the manifest has gone. The ordering you reproduced therefore ends in a refusal and a republish of those packs, instead of a committed inventory entry with no readable snapshot membership.

GC's reasoning is narrowed to match. Below the published head a watermark INSERT can no longer be admitted, because the barrier requires strictly above — so the pair the membership predicate needs can never come into existence through a *new* statement. The case that remains is exactly yours: a statement admitted above the head that stalls and lands below it, and that is what the post-watermark confirmation catches.

To your framing directly: `index_version < published_head` is no longer treated as proof that a manifest can never become relevant. It now carries only the narrower claim about what can still be *admitted*, with the in-flight case handled separately rather than assumed away.

---

tests/tools/verify_replicated_quorum.py
zaoxing: Rewritten in `2bda77f` against your list:

- every branch asserts; the only `FAIL` string left in the file is a docstring describing the old behaviour
- negative branches check the expected exception and the ClickHouse code, rather than accepting any exception
- the success path calls `write_descriptors()` with real `refs`, so descriptor and manifest writes are exercised
- `system.query_log` is required — the script asserts the table exists instead of treating the evidence as optional
- `collect_garbage()` is exercised, which also covers the replicated-mutation path from the GC thread

**Not yet run.** There is no Keeper in this environment, so the rewritten script has not executed once — a rewrite whose branches all assert is worth less than one that has been observed to pass. `docs/catalog-descriptor-key.md` item 2 records the setup (Keeper on 9181, server on 9010) and the expected result: six `PASS` lines and exit 0, with the retention step admitted and removing exactly the orphan rows while both published versions keep theirs. Any `AssertionError` or non-zero exit there is a finding, not noise.

---

src/dmi/storage/capture/clickhouse_lease.py
```diff
@@ -0,0 +1,299 @@
...
+    def renew(self) -> PublisherLease:
+        lease = self._lease
+        if lease is None:
+            raise PublisherLeaseError(
+                "no publisher lease is held; call acquire_publisher_lease() "
+                "before publishing. Only the lease holder can make a snapshot "
+                "visible, and the check rides inside the publish statement, so "
+                "publishing without one writes nothing."
+            )
+        return self.claim(lease.holder, lease_id=lease.lease_id)
```
Samfisheryu, line 78: **Post-merge blocker: one writer can race itself under the same publisher lease.** Two concurrent `publish_snapshot()` calls on one public `ClickHouseCatalogWriter` both renew and reuse this `lease_id`; the server fence therefore treats both operations as the same holder and both return success. With real ClickHouse I observed an already-pinned watermark grow (`W=36: 35→36`, `W=74: 73→74`); in 300 rounds both adjacent watermarks landed every time, and 57 rounds exposed the higher watermark before the lower one arrived. There is no single-thread-only or non-forkable contract, and the design doc explicitly says one writer cannot race itself. Please either serialize one in-flight publish per lease/writer (and document process ownership), or carry a unique publication operation token through renew/fence. Add a regression with two concurrent public calls; this needs a follow-up despite the PR already being merged.

Samfisheryu (commented):
Post-merge correctness finding from real ClickHouse testing. The pack-identity fix is sound, but the publisher lease does not exclude concurrent publishes sharing one writer/lease; details inline.

src/dmi/storage/capture/clickhouse_lease.py
zaoxing: Confirmed, and fixed on `claude/observation-system-architecture-x908yi` (based on the merged `main`, commits `1f1fbcc` tests and `b0c83a9` fix). Traced against the merged code before changing anything:

- `renew()` re-claims with the existing `lease_id`, and `_reject_live` admits a claim when the head has one claimant with that same id, so two concurrent renews on one writer both succeed.
- `fence()` checks `lease_id = mine` and remaining life. It has no notion of an operation, so both publishes pass it.
- The barrier `max(index_version) < V` sits inside each `INSERT ... SELECT`, but ClickHouse does not serialise two such statements against each other, so both evaluate `max()` before either commits. That matches your 300/300 double landing and the 57 rounds where the higher watermark showed first.
- Nothing client-side serialised the calls, and the design doc asserted "one writer cannot race itself" with nothing enforcing it.

**Fix, your first option:** `ClickHouseCatalogWriter` holds one re-entrant lock across `publish_snapshot`, `allocate_version`, `ensure_schema` and every lease operation. One publish is in flight per writer, and therefore per lease; a second caller waits and then publishes above the first, so the barrier admits it in order. The writer also records its owning PID and refuses lease-bearing operations from any other process, because a forked child inherits the parent's `PublisherLease` and would publish under the same `lease_id` from an address space no lock reaches. The contract is now stated in `docs/catalog-descriptor-key.md`: a writer and its lease belong to one process, and threads of that process may share the writer.

I did not take the per-operation token route. It turns the race into a fence-out for one side, but a statement that evaluated its fence before the other operation's renewal still lands afterwards, so it leaves the in-flight window open without adding exclusion the lock does not already give.

**Regression** (`test_two_concurrent_publishes_on_one_writer_are_serialised`): publisher A is held inside a fenced statement while publisher B calls `publish_snapshot` on the same writer; the test asserts B issues nothing, not even its lease renewal. Against the merged writer it fails on that assertion. Also pinned: a failed publish releases the lock, and a writer used from another PID refuses with the parent's lease intact. `make check` and the full `cpu` gate (1115 passed) are green on the fix.

**Not verified here:** the CPU fake cannot model the server-side overlap, so I have not rerun your two-thread, 300-round reproduction on a real ClickHouse. This environment cannot start a server. If you can rerun it against `b0c83a9`, the expected result is that no pinned watermark grows and each round lands the two versions strictly in lock order.

---

PR #125, opened 2026-09-05
Catalog: one writer cannot publish twice at once -- serialise per writer, bind to its process

zaoxing:
Post-merge finding on #119 from real ClickHouse testing: two concurrent
`publish_snapshot()` calls on one `ClickHouseCatalogWriter` both renew the same
`lease_id`, both pass the fence, and the version barrier inside each
`INSERT ... SELECT` is evaluated at admission rather than serialised against
the other statement, so both watermark rows land in either order and a reader
pinned at the higher watermark sees the lower version's membership arrive
underneath it (W=36 grew 35->36, W=74 grew 73->74; 300/300 rounds landed both,
57 exposed the higher first).

The fence identifies a **holder**, not an operation, and ClickHouse offers no
statement-level serialisation to lean on, so the exclusion sits where the
shared identity is minted:

- `ClickHouseCatalogWriter` holds one re-entrant lock across `publish_snapshot`,
  `allocate_version`, `ensure_schema` and the lease operations. One publish is
  in flight per writer, and therefore per lease; a second caller waits. This
  does not order allocation against publication and need not: a thread that
  allocated the lower version and publishes second is refused by the barrier as
  before and re-allocated by the indexer.
- The writer records its owning PID and refuses every lease-bearing operation
  from another process. A forked child inherits the parent's `PublisherLease`
  and would publish under the same `lease_id` from an address space no lock
  reaches.
- Contract, stated in `docs/catalog-descriptor-key.md`: a writer and its lease
  belong to one process, and the writer may be shared between threads of that
  process. A per-operation token through renew and fence was considered and not
  taken; it turns the race into a fence-out for one side but leaves the
  in-flight window open.

Tests first. Against the writer as merged in #119:

    test_two_concurrent_publishes_on_one_writer_are_serialised     FAILED
    test_a_writer_used_from_another_process_refuses_to_publish     FAILED (DID NOT RAISE)
    test_a_publish_that_fails_releases_the_writer_for_the_next_one PASSED

The first fails on its load-bearing assertion: publisher B issued statements
while publisher A's publish was in flight on the same writer. The second fails
because nothing binds a writer to a process. The third passes trivially and
exists to guard the fix. All three pass on this head; the concurrency test was
run 30 times in a row without a flake, and the PID guard was also exercised
against a real `os.fork()` child. The live reproduction (two threads, one
writer, adjacent versions, 300 rounds) is the check that closes it on a real
server; the `clickhouse-live` CI job runs against ClickHouse 25.12.

### DIFF
--- a/src/dmi/storage/capture/clickhouse_catalog.py
+++ b/src/dmi/storage/capture/clickhouse_catalog.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import os
 import secrets
+import threading
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from time import sleep as _sleep, time_ns
@@ -9,6 +11,7 @@ from uuid import uuid4
 from .catalog import (
     CatalogVersionAllocationError,
     PackIdentity,
+    PublisherLeaseError,
     SnapshotPublishConflictError,
     SnapshotPublishRaceError,
 )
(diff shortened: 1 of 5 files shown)


src/dmi/storage/capture/clickhouse_catalog.py
```diff
@@ -764,7 +831,20 @@ def allocate_version(self) -> int:
...
+        not interleave the floor read, the claim and the read-back of one
+        attempt with another's. This does not order allocation against
+        publication -- a thread may allocate the lower version and publish
+        second -- and does not need to: the barrier inside the watermark
+        statement refuses a version that is not above the published head, and
+        the indexer re-allocates. What the lock removes is the case the fence
+        cannot see, two publishes in flight at once under one lease.
         """
+        with self._serial:
+            return self._allocate_version_serialised()
```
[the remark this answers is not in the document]
zaoxing: Correct, and fixed in 13b4731. `allocate_version` now runs `_owned_by_this_process()` under the lock like the other operations, matching the documented contract. The cross-process test asserts it raises `PublisherLeaseError` and that no claim row reached the server; against the previous head that assertion failed with DID NOT RAISE. `make check` is green on the new head.

---

src/dmi/storage/capture/clickhouse_catalog.py
```diff
@@ -305,6 +379,25 @@ def publish_snapshot(
         this module's taxonomy should treat a driver exception from a publish
         as "outcome unknown: re-acquire and re-index".
         """
+        self._owned_by_this_process()
+        with self._serial:
```
Samfisheryu, line 456: **Blocker: quarantine the lease after an outcome-unknown fenced write before releasing this serializer.** The lock covers only the Python call lifetime. If the watermark INSERT has reached ClickHouse but `execute()` returns a transport/interrupt error while the server statement is still running, this context exits with the current lease intact. The waiting publish then renews the same `lease_id` and can publish a higher watermark before the first statement lands. I reproduced this on real ClickHouse: B returned with only W=2 visible; A's delayed W=1 then landed, and the same pinned W=2 membership grew from 1 to 2. Calling `acquire_publisher_lease()` as the doc recommends does not help because an existing local lease deliberately reuses its ID. On an ambiguous outcome, invalidate/quarantine that identity without an early release tombstone, and do not let a successor publish until the old operation is proven finished/cancelled or its lease window has expired. Known-complete errors such as `SnapshotPublishRaceError` can still release normally.

Samfisheryu (changes requested):
The normal completion path is correctly serialized, and the latest commit also closes the shared-client overlap. One outcome-unknown path still recreates the pinned-snapshot race, so I cannot approve this yet.

src/dmi/storage/capture/clickhouse_catalog.py
zaoxing: Fixed in 08f6cb6, as specified: an outcome-unknown failure now invalidates the identity without an early release tombstone, and no successor publishes until the window has expired.

- `ClickHouseLeaseCoordinator.discard_local_lease()` drops the local identity with no tombstone write, so the server row stays live and refuses other writers claims there too.
- `publish_snapshot`, `acquire_publisher_lease` and `renew_publisher_lease` quarantine on any non-taxonomy exception (transport/interrupt included) and refuse for a full `lease_ttl_ns`; known-complete failures (`SnapshotPublishRaceError`, `SnapshotPublishConflictError`, lost lease) release normally and the next publish proceeds at once.
- `release_publisher_lease()` stays a no-op while quarantined, so nothing hands a successor a fresh term inside the window.
- `test_an_outcome_unknown_publish_quarantines_the_lease_until_its_window_expires` pins it: a watermark INSERT that lands and then raises leaves the writer leaseless with no tombstone, refuses every lease-bearing operation, and publishes again only under a fresh identity past the window. The transient-outage fault test now pins refusal-then-recovery instead of immediate convergence.

CPU suites green (348 across catalog/faults/indexer/allocation/reader; full `-m cpu` shows only the 5 pre-existing fsync environment failures that also fail on `main`).

Samfisheryu (approved):
The outcome-unknown path is now closed: the old lease identity is quarantined without an early tombstone, waiting publishers are refused, and recovery uses a fresh identity only after the lease window. The real ClickHouse late-watermark reproduction no longer grows a pinned snapshot; normal 300-round concurrency, shared-client, fork/PID, and recovery regressions also pass.

PR #127, opened 2026-09-06
perf(capture): end-to-end native write path, opt-in via NativePackSink

zaoxing:
## What

Replaces the Python capture-path on the hot side with a C++ implementation
that is byte-identical to the reference. The Python pipeline stays as the
permanent conformance oracle and the rollback target.

```
ring record worker
  → NativePackSink (RecordSink: validate envelopes, ATen dtype map, slice)
  → PackSink: bounded queue → N scope-hashed packer threads
  → per-pair stage queues → stager threads → fsync'd spool
  → parallel uploader → SigV4 PUT (single + multipart) → S3/Garage
```

## Why

| Stage | Python (this host, 5955WX) | Native | Notes |
|---|---:|---:|---|
| Capture pipeline (spool) | 0.235 GiB/s | 0.42–0.52 GiB/s | monotonic to N=8 (+121% vs Python) |
| Packer alone | 0.359 | 0.63 | slice-by-16 IEEE CRC32 |
| Uploader (parallel) | bounded by GIL | bounded by socket | preflight + re-hash + post-upload visibility |

Per-instance: a 3×4090 host needs 1.1 GiB/s; 1.5–2.0 GiB/s modeled ceiling
clears that with 2–3 instances. The design doc's adopted 0.235 target is
beaten with margin in every reading.

> **Read the numbers above as a direction, not a figure.** They were taken
> on a shared host that is not quiet. Measured 2026-09-06: three
> consecutive runs of the same `bench_sink` binary in the same minute, at
> load average 14.5 on 32 threads, gave 0.259 / 0.492 / 0.519 GiB/s — a 2×
> spread on identical work. The *decision* is robust to that: even the
> worst reading beats the 0.235 Python baseline and clears the 0.37 GiB/s
> per-instance requirement for the 3×4090 shape. The *figures* are not
> publishable until they are re-taken on a quiet host, which is the
> outstanding obligation recorded at Checkpoint B and the protocol in
> `docs/benchmarks.md` ("How noisy this host is, measured").

## TDD evidence (every fix went RED first)

Whole-stack CPU suite is 1228 passed / 0 failed / 0 skipped, and the
live suite 188 passed with one pre-existing environmental failure (see
"CI story").
Per-module map — every line in the table has a test in the cited file
that was failing before the implementation went in:

| Module (impl → spec) | Test file | Behavior pinned |
|---|---|---|
| `pack_builder.cpp` (`dmi-pack-v1` writer) | `test_native_pack_conformance.py` | byte-equality vs `PackWriter` on 4 corpora incl. the recorded `tests/data/capture_golden_manifest.json` digest `53a087…`; CRC32 vs zlib; shape/empty/rejection parity |
| `spool.cpp` (`.open`/`.ready`/quarantine) | `test_native_spool.py` | cross-impl drain, idempotence, capacity, quarantine, stale-`.open` removal, accounting, component-wise key validation, symlinked-ancestor containment |
| `pack_sink.cpp` (queue → assembler → spool) | `test_native_pack_sink.py` | golden e2e, scope/linger seals, drop/block policies, duplicates, oversized, size splits, scope isolation N=4, head-to-head vs `HostCapturePipeline` |
| `s3_client.cpp` + `s3_sign.cpp` | `test_native_s3_client.py` + `test_native_s3_sign.py` | 11/11 SigV4 vs botocore, 8/8 against server-side-re-signing fake S3, retry taxonomy, byte cap |
| `uploader.cpp` (parallel + preflight) | `test_native_uploader.py` | e2e sink→spool→store, idempotent re-upload, parallel batch, retries, corrupt refusal, byte-gate, **boto3 head-to-head** |
| `native_pack_sink.cpp` (torch adapter) | `test_native_adapter_torch.py` | 19 tests, all 10 dtypes, validation, lease guards |
| `native_sink.py` (selection) | `test_native_rollback.py` | shared layout, config validation, native→Python indexer e2e |

Live-catalog re-run on rebased origin/main (per the TDD verification
checklist — rebase is a code change, the suite re-runs):

```
tests/test_clickhouse_catalog_live.py + test_clickhouse_capture_catalog.py
                                  + test_capture_end_to_end_live.py
-m "clickhouse and manual and not garage" -q
  → 11 passed, 100 deselected in 7.19s
```

This includes the two new tests that drove #125 (per-writer publish
serialization, process binding) — Phase A's native write path never
touches the catalog, but the rollback test (above) uses the Python
`CatalogIndexer`, so the catalog fix lands cleanly underneath Phase A
and Phase B (the native catalog port) inherits the new contract.

Full suite after rebase:
```
make test    → 1228 passed, 149 deselected
make test-all → 1207 passed, 7 pre-existing failures (missing
                _native_backend/_host_backend .so + uninitialized
                submodules), identical with or without this branch
```

## Design choices, in ledger

`docs/benchmarks.md` records every kept idea and every dead one:
- raw-FNV routing silently collapsed all 8 scopes onto worker 0 (flat
  "scaling" — a real bug); fmix64 finalizer is the fix
- broadcast `notify_all` on every submit was a 180k-wakeup per-trial
  thundering herd at N=8; per-worker + per-stage CVs fixed scaling
- slice-by-16 IEEE CRC32 (not crc32q) — same polynomial, far faster
- Lock scope in `Spool::Stage` covers accounting only; byte reservation
  keeps the byte cap exact
- `.dmi-pack.ready` suffix check was 15 chars when it should be 16 →
  every ready file invisible to recovery; the test caught it
- `Close()` calling `Snapshot()` under the held lock self-deadlocked
  (the test caught it on the first `native-build conformance_sink` run)
- Forward-wait used the wrong CV → flush barrier hung; test caught it
- Fake S3 emitted duplicate `Content-Length` on HEAD → boto3 head-to-
  head test caught it; S3-faithfulness improved for every client test
- `".."` was tested as a *substring* of the whole object key, not as a
  path component, so the legal key `tenant=a..b` was refused — and the
  refusal fails the stager thread, which closes the sink for every
  later submit. Python stages that key fine. Now a component-wise walk
  against Python's `_KEY_COMPONENT`, which also *narrows* native for
  keys carrying other illegal characters
- The escape guard was purely textual, so a symlinked ancestor let a
  pack be written *outside* the spool root and acknowledged as durably
  staged — then `recover()` could not see it, because
  `recursive_directory_iterator` does not follow directory symlinks.
  The pack was never uploaded and its bytes stayed charged against the
  budget forever. Now `weakly_canonical` containment before the lock,
  matching the oracle's `resolve(strict=False)`

## Post-review polish round (the last 24 commits)

A four-lens review of `native/csrc/` (correctness, consistency,
test-coverage, simplification) reported 40 findings across the stack.
Each was handed to an independent adversarial verifier whose job was to
*refute* it; only verifier-confirmed `correctness`/`requirement`
findings were fixed. **3 were refuted outright and 6 more were confirmed
but downgraded to `optional`** — those are listed below rather than
fixed, so a reviewer can tell "judged" from "missed".

Because #128 and #129 merged mid-review, these fixes were written
against the three separate branches and then replayed onto this one.
The conflicts were resolved to keep BOTH sides wherever both carried a
real check — see the `test_the_inline_identity_list_is_byte_identical_too`
/ `test_the_parameterized_statements_are_byte_identical_too` pair, which
were two same-named tests covering different paths (the inline member
list vs the `%(name)s` parameter path); both survive, renamed.

What landed, by area:

| Area | Fixes |
|---|---|
| `spool.cpp` | `a..b` component check (a legal key latched the whole sink); `weakly_canonical` containment (a symlinked ancestor wrote outside the root and lied about durability) |
| `common/json.cpp` | the shared integer scan is bounded before the multiply, and the row path refuses an out-of-range literal instead of packing its remainder |
| `catalog_writer.cpp` / `clickhouse_client.cpp` | one ten-character escaper (was four of ten); `CURLOPT_POSTFIELDSIZE_LARGE` (a NUL truncated the statement on the wire); single-pass `%(name)s` substitution (a value spelling a placeholder was rewritten by a later parameter's pass) |
| `pack_index.cpp` | footer metadata validated where the oracle validates it, with overflow-checked shape and offset arithmetic; `pack_id` validated as a UUID instead of interpolated raw |
| `schema.cpp` | the rebuild instruction now names the legacy object it refuses over; the grant probe covers it too |
| `indexer.cpp` | one failure per distinct claimant, in the oracle's order |
| `reader.cpp` | cursor `captured_at_ns`, `v` and `w` all bounded (each wrapped modulo 2^64); leading zeros refused; the four missing `CaptureQuery` bounds and the oracle's text validation ported |
| `hydration.cpp` | rank ≥ 2 hydration and the whole-record footer binding pinned by tests |

Every correctness fix went RED first. As one worked example, I re-proved
the two `spool.cpp` fixes independently of their author by reverting
only that file to `92c2225`, rebuilding the drivers, and re-running:

```
FAILED test_native_spool.py::test_stage_rejects_a_key_escaping_through_a_symlinked_directory
FAILED test_native_pack_sink.py::test_dotted_tenant_stages_and_does_not_latch_the_sink
2 failed, 1 passed, 20 deselected
```

The test that passes in both states is the point:
`test_object_key_parity_with_python` only ever compared the key
*string* and never staged, so a defect that latches the whole sink was
invisible to the test named for object-key parity. It now stages, and
asserts a later submit is still accepted.

### Confirmed on this PR's files but deliberately NOT fixed

Recorded so a reviewer does not re-find them. Each was confirmed by
execution and then judged not worth a behavior change here:

| Finding | Why it is not fixed |
|---|---|
| `s3_sign.cpp:75` sorts `SignedHeaders` by the raw map key while lowercasing only the emitted text | Every production path lowercases before signing, so it is reachable only through the conformance driver. Verified: mixed-case `X-Amz-Meta-Zed` diverges from botocore, `Content-MD5` does not |
| `pack_builder.cpp:353` discards `ParseUuid`'s result | Unreachable: the only non-test caller feeds `NewPackId()`, and `SinkConfig` has no `pack_id` field, so there is no injection surface. Python's reader also rejects such bytes on read |
| `spool.cpp:34` `IsUuid` accepts uppercase hex where Python's ready-name grammar is lowercase-only | No in-tree producer can emit one. Worth knowing that where it *is* reachable Python **quarantines** the file rather than skipping it |
| `pack_sink.cpp:64` validates only `num_workers` | The contract is deliberately enforced one layer up, in `native_sink.py`, whose docstring says so. 3 of the 5 bounds diverge, not "every other bound" |
| `uploader.cpp:221` records an oversized pack per-pack where the oracle refuses the whole batch up front | **Open design call, not a defect to fix silently.** Measured: Python raises with 0 objects uploaded and 4 packs still staged; native returns `ok:true` with 3 uploaded and 1 staged. Two purpose-named tests pin the current behavior and `docs/benchmarks.md:483` records a related decision, so choosing the oracle's all-or-nothing policy is a maintainer call. What is unambiguously wrong either way is the comment at `uploader.cpp:222` asserting the parity it violates |
| `spool.cpp:277` conflict check outside the write lock | **Refuted.** The TOCTOU is real but both claimed harms were disproved by execution: `::link` returns `EEXIST` rather than overwriting, and the loser's `unreserve()` keeps accounting exact. Do not "fix" this by moving the write under the mutex — it would serialise every stager on disk I/O and still not close the cross-process case |

## What is *not* in this PR (honest scope)

- No default-sink change: `create_native_pack_sink` is opt-in via
  `MonitoringEngine.create_record_runtime(record_sink=...)`. The Python
  path is unchanged for everyone else.
- ~~Phase B (native catalog/indexer) is the next step~~ — **no longer
  true.** #129 merged into #128 and #128 merged into this branch, so
  this PR now carries all three phases: the native write path (A), the
  native catalog and indexer (B), and the native reader and hydration
  (C). Read the diff as the whole stack, not as Phase A alone.
- **Live ring e2e**: the adapter is driven with synthetic envelopes.
  A real `RingEngine` path requires the full backend (submodule-init
  blocked here). Rollback proven via the Python-side indexer instead.
- **Live Garage**: all store tests run against the signature-verifying
  fake, not a real endpoint. Boto3 head-to-head runs against the fake
  too (URL-rewritten via the test's proxy slot).

## CI story

- `make test` (CPU): 1228 passed.
- `make test-all` (full): 7 pre-existing failures unrelated to this
  work — missing build artifacts (gitignored `.so` for `_native_backend`
  / `_host_backend`, submodules uninitialized here). Same 7 with or
  without this branch.
- The S3/uploader tests skip gracefully when libcurl headers are
  absent; the torch adapter needs the venv torch CI installs
  (`torch>=2.8,<3` is in `python-checks.yml`).

## Files (89 changed, +24 436)

- `native/csrc/{common,pack,store,sink}/**` — the implementation
- `native/csrc/conformance/` — moved into `common/`
- `tests/test_native_{pack_conformance,s3_sign,s3_client,spool,pack_sink,uploader,adapter_torch,rollback}.py` — TDD tests, all RED first
- `benchmarks/profile_pipeline_stages.py`, `benchmarks/data/native-pipeline/{baseline-*,}.json` — attribution
- `docs/benchmarks.md` — perf ledger (every kept and reverted decision)
- `tasks/{plan,todo}.md` — plan and durable state (re-read at every cycle)

## Reviewer's quick path

1. `make test PYTHON=<venv>/bin/python` — 1228 passed, 0 failed.
2. `pytest -m 'clickhouse and manual and not garage' PYTHON=<venv>/bin/python`
   — live-catalog pass (11, includes #125's two new tests).
3. `native/build/conformance_{main,sign,store,spool,sink}` (CPU-only
   drivers) cover the byte-compatibility, signing, and fault paths
   without needing a GPU or ClickHouse.
4. `native/build/bench_sink single` and
   `native/build/bench_sink interleaved spooldir=/dev/shm/...` reproduce
   the scaling numbers.


native/csrc/store/uploader.cpp
```diff
@@ -0,0 +1,343 @@
...
+      result.failures.push_back(
+          {staged.pack_id, staged.object_key, 0,
+           "pack exceeds the in-flight byte limit"});
+      ++result.snapshot.attempted_packs;
+      ++result.snapshot.failed_packs;
+      continue;
+    }
+  }
+  result.refs.resize(pending.size());
+  result.failures.resize(pending.size());
```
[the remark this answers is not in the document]
zaoxing: Fixed in d207668: `UploadPending` now assigns both vectors to the recover() order up front and writes oversized refusals into their own indexed slots; workers fill the rest by index, so refs[i]/failures[i] always pair positionally. The new mixed-batch test (`test_mixed_batch_reports_oversized_pack_at_its_position`) reproduced the loss first — before the fix all four failure records came back empty while the snapshot still said failed_packs=1 — and passes after.

native/csrc/pack/conformance_main.cpp
```diff
@@ -0,0 +1,209 @@
...
+      meta.batch_position = static_cast<uint64_t>(meta_int(rt, "batch_position"));
+      meta.dtype = meta_string(rt, "dtype");
+      meta.shape = meta_shape(rt);
+      meta.captured_at_ns = static_cast<uint64_t>(meta_int(rt, "captured_at_ns"));
+      // payload
+      static std::vector<std::vector<uint8_t>> payload_storage;
+      payload_storage.push_back(meta_payload(rt));
+      rec.payload = payload_storage.back().data();
+      rec.payload_bytes = payload_storage.back().size();
+      const dmi_pack::Status st = builder.Append(rec);
```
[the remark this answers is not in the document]
zaoxing: Fixed in d207668: the static vector is gone. `PackBuilder::Append` copies the payload into its buffer synchronously, so a per-record local has the right lifetime; the pack conformance suite (incl. golden byte-equality) is green against the rebuilt driver.

tests/test_native_pack_conformance.py
```diff
@@ -0,0 +1,381 @@
...
+        step_number=0,
+        token_start=0,
+        token_end=0,
+        batch_position=0,
+        dtype="uint8",
+        shape=(2, 3),
+        captured_at_ns=1,
+    )
+    records = [CaptureRecord(metadata=CaptureMetadata(**meta), payload=bytes(6))]
+
```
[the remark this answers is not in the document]
zaoxing: Fixed in d207668: the test now calls `_assert_packs_match` like its empty-shape sibling, so multi-dim shape handling is actually compared byte-for-byte. Verified it has teeth by mutating the payload (test fails) and restoring it (passes).

native/csrc/common/json.cpp
```diff
@@ -0,0 +1,279 @@
...
+          out->append(buf);
+        } else {
+          out->push_back(static_cast<char>(c));
+        }
+    }
+  }
+  out->push_back('"');
+}
+
+}  // namespace dmi_conformance
```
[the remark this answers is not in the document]
zaoxing: Fixed in d207668 — closing comment now reads dmi_common.

native/csrc/common/json.h
```diff
@@ -0,0 +1,62 @@
...
+
+// A "..." literal (with quotes) to raw text.
+std::string ParseLiteral(const std::string& literal);
+
+bool DecodeBase64(const std::string& in, std::vector<uint8_t>* out);
+void EncodeBase64(const std::vector<uint8_t>& data, std::string* out);
+void EncodeBase64(const uint8_t* data, size_t n, std::string* out);
+void EscapeJson(const std::string& value, std::string* out);
+
+}  // namespace dmi_conformance
```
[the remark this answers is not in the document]
zaoxing: Fixed in d207668 — closing comment now reads dmi_common.

tests/test_native_spool.py
```diff
@@ -0,0 +1,219 @@
...
+# The key must end in "<pack_id>.dmi-pack" — both spools enforce this — and
+# its tenant= segment must name the pack's own tenant (the catalog's
+# foreign-tenant binding refuses anything else at read time).
+OBJECT_KEY = (
+    "v1/tenant=tenant-golden/date=2026-09-05/session=session-golden/rank=0/"
+    "018f0000-0000-7000-8000-000000000f01.dmi-pack"
+)
+
+
+def _golden_pack() -> bytes:
```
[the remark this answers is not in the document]
zaoxing: Fixed in d207668 — annotation corrected to `tuple[bytes, str, int]`.

zaoxing:
Copilot's bench_sink thread (spool cleanup via `system(("rm -rf ") + argv)`) — **reproduced** (`spooldir=/tmp/x; touch /tmp/marker` executed the marker) and **fixed in 925de3d** on this branch's stack: the cleanup is `std::filesystem::remove_all` on the path itself, no shell, no injection. CPU gate 1203 green after. Thread resolved.

Samfisheryu (commented):
Reviewed this PR together with the #128/#129 stack; four reproduced findings are inline. The Ring binding and sustained spool/upload lifecycle need fixing before merging the native path.

Validation at #129 head a4d3dcb (findings attributed to the introducing diff): native pack/store/sink suites 78 passed; real ClickHouse catalog/reader suites 66 passed; existing single-GPU reference/ClickHouse record Ring suites 10 passed. However, the actual default native capture runtime fails during extension loading, and the full CPU suite cannot collect when both extensions are built. Excluding the two conflicting test modules, the remaining CPU suite passed 1181 tests. S3 tests used the signature-verifying fake service, not live Garage. No GPU Ring implementation changes or hypothetical multi-stream ordering assumptions are involved.

native/csrc/sink/bindings_sink.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+}  // namespace
+
+PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
+  py::class_<ring::RecordSinkLease, std::shared_ptr<ring::RecordSinkLease>>(
+      m, "RecordSinkLease");
+
+  py::class_<dmi_sink::NativePackSink,
+             std::shared_ptr<dmi_sink::NativePackSink>>(
+      m, "NativePackSink")
```
Samfisheryu, line 112: [P1] Reuse the existing Ring binding types so this sink can actually attach

With both extensions rebuilt, this minimal import fails:

```python
import torch
import _native_backend
import _dmi_native_sink
# ImportError: generic_type: type "RecordSinkLease" is already registered!
```

`bindings.cpp` already registers this same C++ lease type. I also initialized a real single-GPU Ring and called #129's default `create_record_runtime(CaptureRecordFormat())`; it fails at the same import, before any records can flow.

There is a second part to the binding fix: this `NativePackSink` registration omits `ring::RecordSink` as its pybind base, so eliminating the duplicate registration alone does not provide the engine's required type relationship or inherited `_acquire_engine()` method. Please reuse the existing lease/base registration and bind the derived sink correctly. Add a test loading both extensions and attaching this sink to a real Ring; the standalone synthetic-envelope tests do not exercise that boundary.

src/dmi/storage/capture/native_sink.py
```diff
@@ -0,0 +1,118 @@
...
+        module = _load_native_sink_extension()
+        self._native_sink = module.NativePackSink(
+            spool_root=config.spool_root,
+            layout=LAYOUT_NAME,
+            num_workers=config.num_workers,
+            max_queue_records=config.max_queue_records,
+            max_queue_bytes=config.max_queue_bytes,
+            max_pack_bytes=config.max_pack_bytes,
+            max_pack_records=config.max_pack_records,
+            max_linger_ns=config.max_linger_ns,
```
Samfisheryu, line 129: [P2] Forward spool_max_bytes through the factory and binding

The config validates `spool_max_bytes`, but this constructor call does not pass it, and the extension constructor has no corresponding argument. `SinkConfig` therefore retains its 1 TiB default.

Reproduced with `NativeSinkConfig(spool_root=..., spool_max_bytes=1)`: attach the resulting sink using the existing adapter test helper, submit one 16-element float32 row (64 payload bytes), then flush. Flush succeeds and leaves an **854-byte ready pack**, despite the one-byte limit.

Please carry the value through Python config → pybind constructor → `SinkConfig::spool_max_bytes`, and test the actual configured capacity refusal.

native/csrc/store/spool.cpp
```diff
@@ -0,0 +1,514 @@
...
+          return SpoolStatus::kConflict;
+        }
+      }
+      if (bytes_ + n > max_bytes_) {
+        if (error) {
+          *error = "spool byte limit exceeded: " +
+                   std::to_string(bytes_ + n) + " > " +
+                   std::to_string(max_bytes_);
+        }
+        return SpoolStatus::kFull;
```
Samfisheryu: [P1] Let the running sink reclaim capacity released by the uploader

`PackSink` owns a private `Spool` with no shared spool handle exposed to the uploader. Opening the same root for upload/removal updates that second object's counters, not the writer's `bytes_`. The writer eventually reports full even after all staged files have been uploaded and removed.

Reproduced **in one process, completely serially**: configure `PackSink` with a 1500-byte spool limit; submit/flush a 16-element float32 record; open the uploader-side `Spool` on that root and `Recover`/`Remove` the ready pack; then submit/flush a second distinct record through the still-running sink:

```text
round=0 removed=1 remaining bytes=0
round=1 flush failed: spool stage failed: spool byte limit exceeded: 1700 > 1500
```

The directory is empty at the second failure. The equivalent sink → fake-S3 uploader cycle also reproduces it. This does not require multiple uploaders or cross-process scheduling. Please expose/share a single spool ownership/accounting boundary, or otherwise reconcile released capacity before admission.

native/csrc/store/uploader.cpp
```diff
@@ -0,0 +1,343 @@
...
+          ref->pack_id = staged.pack_id;
+          ref->store_id = config_.store_id;
+          ref->object_key = key;
+          ref->object_bytes = staged.object_bytes;
+          ref->checksum = staged.checksum;
+          ref->record_count = staged.record_count;
+          if (attempts_out) *attempts_out = attempts;
+          return true;
+        }
+      }
```
Samfisheryu, line 160: [P1] Refuse an existing conflicting object instead of overwriting it

When HEAD finds the key but its size/checksum does not match, this branch falls through to the subsequent PUT. That replaces the existing object, then removes the local ready pack and reports success.

Repro with the existing fake-S3 uploader fixture: stage one pack; prepopulate `staged['object_key']` with `b'previous immutable object'` and different metadata; call `UploadPending`. Observed: `uploaded_packs=1`, `failed_packs=0`, the original object body is replaced by the new pack, and the ready file is gone. The Python store's equivalent operation rejects this as a pack conflict.

Please return a conflict and retain the local pack whenever the already-existing object differs. This is not asking for cross-writer CAS: HEAD has already detected the conflicting object before this PUT.

native/csrc/sink/bindings_sink.cpp
Samfisheryu: Still reproducible at 8748b91, including the explicit entry point now advertised in the PR body. Rebuilt the native sink extension and initialized a real single-GPU Ring (64 KiB payload/staging, 32 tasks):

```python
handle = create_native_pack_sink(NativeSinkConfig(spool_root=root))
isinstance(handle.native_sink, RecordSink)           # False
hasattr(handle.native_sink, "_acquire_engine")       # False
engine.create_record_runtime(
    CaptureRecordFormat(), record_sink=handle.native_sink
)
# TypeError: record_sink must be a native RecordSink
```

The automatic capture_sink_config entry fails identically. Module-local registration removed the duplicate-import error, but did not preserve the main backend's Python base class or engine-lease method. Please reuse the existing Ring binding relationship rather than adding a second Python RecordSink type, and test the real engine attachment.

The old reference/ClickHouse GPU E2E suites pass (10 tests); the new path fails before records flow. This is an adapter/binding issue, not a proposed change to the GPU Ring design.

native/csrc/store/spool.cpp
```diff
@@ -0,0 +1,596 @@
...
+        uint64_t ready_count = 0;
+        for (const auto& entry :
+             fs::recursive_directory_iterator(root_, walk_ec)) {
+          if (walk_ec) break;
+          if (!entry.is_regular_file()) continue;
+          const std::string name = entry.path().filename().string();
+          if (HasSuffix(name, kReadySuffix)) ++ready_count;
+        }
+        bytes_ = actual;
+        entries_ = ready_count;
```
Samfisheryu: [P1] Keep in-flight reservations when reconciling spool capacity

The serial uploader-removal repro is fixed, but replacing the counters with the directory snapshot loses reservations that another Stage has made without finishing its file write.

Deterministic repro on 8748b91: one process, one Spool, max_bytes=1500. Thread A reserves 1000 bytes; pause it just before opening its .open file. Thread B stages another 1000 bytes, triggering this scan. Then resume A:

```text
A=ok B=ok actual_bytes=2000 reported_bytes=1000 limit=1500
```

I reproduced this by wrapping open with the linker's --wrap=open, not by changing production code. The same probe against the pre-reconciliation spool.cpp rejects B. This is ordinary CPU stager concurrency, unrelated to GPU streams or TP.

Please account for committed files and in-flight reservations separately so a scan cannot erase the latter. Keep both the successful serial upload/remove case and this concurrent admission case as regressions.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,892 @@
...
+  int64_t footer_requests = 0;
+  int64_t footer_bytes = 0;
+  const int64_t payload_requests_total =
+      static_cast<int64_t>(payload_requests);
+  for (const auto& plan : plans) {
+    const int64_t remaining_requests = request_limit - payload_requests_total;
+    const int64_t remaining_bytes =
+        byte_limit - static_cast<int64_t>(payload_bytes);
+    if (footer_requests + 2 > remaining_requests ||
+        footer_bytes > remaining_bytes) {
```
Samfisheryu: [P1] Enforce the footer byte budget before fetching it

The request-count half of the earlier [budget finding](https://github.com/ProjectDMX/DMI/pull/129#discussion_r3946402572) is fixed, but the byte-limit half still fails on 8748b91.

Repro: stage/upload/index/select one uint8 tensor with shape=(16,) and payload=bytes(range(16)). Call hydrate with byte_limit=16 and request_limit=3. Python refuses; native makes all three GETs (trailer, footer, payload) and returns:

```json
{"ok":true,"payloads":["AAECAwQFBgcICQoLDA0ODw=="]}
```

Here footer_bytes is checked before the current pack is charged. For a one-pack selection there is no next iteration to reject the overrun. The later accounting also estimates using object_bytes instead of the actual footer range length.

Please charge/check each actual range before fetching it, including the last pack's footer. Raising the request limit from 1 to 3 must not bypass byte_limit.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,892 @@
...
+        std::string current;
+        bool in_str = false;
+        int depth = 0;
+        for (size_t i = 0; i < rendered.size(); ++i) {
+            const char c = rendered[i];
+            if (in_str) {
+                if (c == '\\' && i + 1 < rendered.size()) {
+                    current.push_back('\\');
+                    current.push_back(rendered[++i]);
+                    continue;
```
Samfisheryu: [P2] Compare decoded metadata, not SQL-escaped footer text

The full-field comparison now rejects valid captures. Reproduced on 8748b91 with three otherwise identical records whose hook_name is respectively r"block\resid", "block'quoted", and "block\tresid" (a tab). Stage/upload/index/select succeeds; Python hydrate returns the original 16-byte payload; native hydrate fails in all three cases:

```text
catalog descriptor does not match the pack footer: field 12 for upload-0001
```

This branch preserves the footer renderer's backslash escape, while the catalog field is already decoded. Comparing these different representations produces a false mismatch.

Please decode the SQL string representation before comparison, or compare typed descriptors directly. Keep the new full-field identity check; the fix is consistent representation, not dropping fields again.

Samfisheryu (commented):
Re-review of the complete #127 stack at 8748b91 (now including #128/#129).

Confirmed fixed: cross-parameter SQL substitution, cursor timestamp/filter bounds, and the original shape/dtype/conflict/limit regressions. Verification: 1,427 CPU tests, 129 live-ClickHouse tests, and 10 legacy reference/ClickHouse single-GPU Ring E2E tests passed.

Not ready to approve yet: the explicit native sink still fails the real Ring RecordSink check (follow-up on the existing inline), and the inline findings below reproduce capacity, hydration and metadata correctness gaps. Synthetic-envelope tests do not establish live Ring E2E. S3 tests used the signature-checking fake, not live Garage.

One scope clarification: the PR body says explicit record_sink only, but engine.py still auto-creates the native sink when storage_backend="capture" and capture_sink_config is provided. The legacy ClickHouse path is not replaced; its E2E tests passed. Please align the description with the intended selection behavior.

The documented INT64_MIN abs_max_int saturation is an intentional contract choice, not counted here as an unacknowledged bug.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,892 @@
...
+
+// Normalize one field for comparison. Two form differences between the
+// rendered footer and the catalog's TSV rows:
+//   1. The footer renders a NULL metadata value as the unquoted three-
+//      letter text, while the catalog rows carry the empty string (TSV \N).
+//   2. The footer renders UUID-typed columns as toUUID('...') — SQL the
+//      INSERT path needs — while the catalog TSV carries the bare UUID
+//      string.
+std::string normalize_footer_field(const std::string& value) {
+  if (value == "NULL") return std::string();
```
Samfisheryu: [P2] Preserve the distinction between SQL NULL and the string "NULL"

This content-based normalization is applied to both the footer and catalog fields, so it can accept metadata that disagrees.

Repro on 8748b91: write/index a valid pack with adapter_revision="NULL"; then, only in the isolated test catalog, change that field to SQL NULL:

```sql
ALTER TABLE <test_prefix>_capture_raw
UPDATE adapter_revision = NULL WHERE 1 SETTINGS mutations_sync=2;
```

Select and hydrate again. Python refuses the descriptor/footer mismatch; native returns ok=true and the payload. The stored pack still contains the literal string "NULL".

Please retain nullness separately from string contents. An ordinary quoted string must not be normalized to the same value as an absent field.

native/csrc/sink/record_row.cpp
```diff
@@ -0,0 +1,198 @@
...
+  out->layer_number = layer;
+  out->producer_rank = static_cast<uint64_t>(producer);
+  out->step_number = static_cast<uint64_t>(step);
+  out->token_start = static_cast<uint64_t>(token_start);
+  out->token_end = static_cast<uint64_t>(token_end);
+  out->batch_position = static_cast<uint64_t>(batch);
+  out->dtype = jc::FindString(text, "dtype");
+  out->shape.clear();
+  for (const auto& item :
+       jc::SplitElements(jc::Unwrap(jc::FindArray(text, "shape")))) {
```
Samfisheryu: [P2] Complete scalar support in both the sink parser and summary

The pack-index scalar fix does not cover this actual torch sink boundary. Repro using the existing adapter-test helpers:

```python
meta = _meta(0)
meta["shape"] = []
CaptureMetadata.from_mapping(meta)  # valid scalar
sink, lease = _make_sink(tmp_path)
sink.submit_envelope(
    LAYOUT, [_row(meta, 0, 4, "float32", shape=())],
    torch.tensor(3.5, dtype=torch.float32),
)
# RuntimeError: NativePackSink: invalid capture metadata:
# capture shape must be an integer list
```

SplitElements(Unwrap("[]")) produces one empty item rather than zero dimensions. Scalar shape=[] must mean one element.

There is a second copy to fix in [hydration.cpp::shape_product](https://github.com/ProjectDMX/DMI/blob/8748b9154e5ff9f9e0eada3a51e6caae6793b58a/native/csrc/catalog/hydration.cpp#L299-L305): staging the same scalar through the pack builder now permits indexing/hydration, but summarize_core still fails with "payload length does not match dtype and shape: 4". Please pin both sink admission and scalar summary, not just scalar indexing.

native/csrc/sink/record_row.cpp
```diff
@@ -0,0 +1,198 @@
...
+  }
+  if (not_an_integer) {
+    return fail("capture metadata field is not an integer");
+  }
+  out->layer_number = layer;
+  out->producer_rank = static_cast<uint64_t>(producer);
+  out->step_number = static_cast<uint64_t>(step);
+  out->token_start = static_cast<uint64_t>(token_start);
+  out->token_end = static_cast<uint64_t>(token_end);
+  out->batch_position = static_cast<uint64_t>(batch);
```
Samfisheryu, line 118: [P2] Reject invalid numeric metadata instead of silently converting it

The overflow/null fixes still leave numeric values that Python CaptureMetadata rejects but this sink accepts and persists. Reproduced against the rebuilt torch extension using _meta/_row/_make_sink from test_native_adapter_torch.py, then flush and read the staged pack with Python PackReader:

| Input metadata | Persisted value |
| --- | --- |
| layer_number=0.5 | 0 |
| captured_at_ns=123.5 | 123 |
| step_number=-1 | 18446744073709551615 |
| shape=[16.5], envelope shape=[16] | [16] |

The [shared integer scanner](https://github.com/ProjectDMX/DMI/blob/8748b9154e5ff9f9e0eada3a51e6caae6793b58a/native/csrc/common/json.cpp#L84-L100) accepts a digit prefix without requiring the complete token to be an integer; the shape loop does likewise. Separately, these casts turn negative counters into valid-looking UInt64 values.

Please require complete integer tokens and reject negative unsigned counters before conversion. The standard CaptureRecordFormat validates metadata upstream, so this is a native sink input-contract gap, not a claim that normal GPU-generated metadata always corrupts.

native/csrc/common/json.cpp
```diff
@@ -0,0 +1,316 @@
...
+          q += 5;
+          if (v < 0x80) {
+            raw.push_back(static_cast<char>(v));
+          } else if (v < 0x800) {
+            raw.push_back(static_cast<char>(0xC0 | (v >> 6)));
+            raw.push_back(static_cast<char>(0x80 | (v & 0x3F)));
+          } else {
+            raw.push_back(static_cast<char>(0xE0 | (v >> 12)));
+            raw.push_back(static_cast<char>(0x80 | ((v >> 6) & 0x3F)));
+            raw.push_back(static_cast<char>(0x80 | (v & 0x3F)));
```
Samfisheryu, line 57: [P2] Decode a JSON surrogate pair as one Unicode code point

A valid non-BMP identifier is corrupted by the native indexer. Repro on 8748b91: use hook_name="block😀" in the existing sink→upload→index flow. Python PackReader reads the original hook name correctly from the uploaded pack, but SELECT hex(hook_name) from the test catalog gives:

```text
expected: 626C6F636BF09F9880
actual:   626C6F636BEDA0BDEDB880
```

Indexing reports success; Python CaptureReader.search then raises UnicodeDecodeError on byte 0xed. The BMP control "block中文" completes the same round trip.

The canonical footer contains a surrogate pair (\\ud83d\\ude00). This routine encodes each half separately into three bytes rather than combining them into the single four-byte UTF-8 code point. Please combine valid high/low surrogate pairs when decoding. This is verified metadata-byte corruption, not a tensor-payload mismatch.

native/csrc/sink/bindings_sink.cpp
zaoxing: Fixed in 96c01e4.

`_dmi_native_sink` no longer registers a `RecordSink`/`RecordSinkLease` of its own. At module init it loads the main backend through `dmi.transport.native._load_extension()` (the same file-path loader the engine uses) and then registers `NativePackSink` with `ring::RecordSink` as its pybind base, which pybind11 resolves to the main module's registration through the shared registry. So `NativePackSink` is a Python subclass of `_native_backend.RecordSink`, `isinstance(handle.native_sink, RecordSink)` is True and `_acquire_engine` is inherited, not re-implemented. `native_sink.py` also loads the main backend before the sink extension. Only when the main backend is not built at all (torch-CPU test hosts) does the module fall back to module-local stand-ins, and it says so via `RING_TYPES_ARE_STANDINS`.

Test: new `tests/test_native_sink_ring_e2e.py` (gpu marker) loads both extensions, asserts the base-class relationship, attaches the sink through `create_record_runtime(..., record_sink=handle.native_sink)` on a real single-GPU Ring (64 KiB payload/staging, 32 tasks), pushes a Ring record, a CPU-direct record and a scalar, and reads them back from the staged pack; a third case takes the automatic `storage_backend="capture"` + `capture_sink_config` entry point. `test_native_adapter_torch.py` also asserts the MRO when the main backend is present. I do not have a CUDA host here, so the GPU file is written against the existing e2e scaffolding but was not executed locally.

native/csrc/store/spool.cpp
zaoxing: Fixed in 96c01e4.

`Spool` now keeps two accounts: `committed_bytes_/entries_` (ready files on disk, what a scan can re-derive) and `reserved_bytes_/entries_` (fresh stages between their capacity check and their `link()`). Capacity is judged against the sum. The reconciling scan overwrites only the committed account; reservations are never touched by it, and this object's own in-flight `.open` temp files are named in a set so the scan does not double count them. `Recover()` likewise rebuilds only the committed account.

Both cases are pinned in `tests/native/test_spool_reservations.cpp` (run by `tests/test_native_spool_reservations.py`, CPU marker): the serial upload/remove through a second Spool object still succeeds, and your concurrent case (A reserves 1000 and is held before its write via a new `SetStageHookForTesting` seam; B stages 1000 against max 1500) now gives B `kFull`, A `ok`, 1000 bytes on disk and `Snapshot().bytes == 1000`. The existing spool/sink/uploader suites pass unchanged.

native/csrc/catalog/hydration.cpp
zaoxing: Fixed in 96c01e4.

`read_pack_descriptor_rows` takes a `charge(length)` hook and calls it before each GET: 64 bytes for the trailer, then the footer at the length the trailer declares. `hydrate` passes a `ReadBudget` (requests and bytes remaining after the payload ranges, the same shape as Python's `_ReadBudget.consume`) so every range, including the last pack's footer, is checked before it is fetched, and at its actual length rather than `object_bytes`. Your repro (`byte_limit=16, request_limit=3`, one 16-byte uint8 capture) is refused with `hydration byte limit exceeded` before any GET.

Tests: `test_the_footer_byte_budget_is_charged_before_the_footer_is_fetched` in `test_native_reader_parity_live.py` (byte refusal, request refusal at `request_limit=2`, and the granted control at `request_limit=3` with room for the bytes), Python as the oracle. That file needs a live ClickHouse, which I do not have here, so it was written against the existing `_e2e_setup` scaffolding but not run locally.

native/csrc/catalog/hydration.cpp
zaoxing: Fixed in 96c01e4.

The footer row is now split into typed fields (`split_footer_row`): a quoted literal is decoded through the inverse of `sql_quote` (all ten escapes), `toUUID('...')` is unwrapped, arrays split on top-level commas only. The comparison (`footer_field_matches`) is decoded-text against the catalog's decoded text, so `block\resid`, `block'quoted` and the tab-bearing name bind. The full-field identity check is kept as is.

Tests: the decoder is pinned on the CPU gate through a new session-less driver op `footer_row_fields` in `tests/test_native_hydration_footer_fields.py` (renders through the driver's own `escape` op, so it tracks the real escaper), and the live repro is `test_hydrate_accepts_metadata_the_sql_escaper_rewrites` (parametrised over the three names, payload compared against Python's hydrate) in `test_native_reader_parity_live.py`. The live test was not run locally (no ClickHouse here).

native/csrc/catalog/hydration.cpp
zaoxing: Fixed in 96c01e4.

Nullness is now a field kind, not a spelling: only the unquoted `NULL` token in the footer row is null, and it matches the catalog's empty (TSV `\N`) field and nothing else. A quoted `'NULL'` decodes to the four-letter string and is compared as a string, so it matches a catalog `"NULL"` and does NOT match a catalog SQL NULL. The content-based normalisation is gone.

Tests: `test_sql_null_and_the_string_null_stay_apart` in the new CPU-gate `tests/test_native_hydration_footer_fields.py` covers all four combinations, and `test_hydrate_keeps_sql_null_apart_from_the_string_null` in `test_native_reader_parity_live.py` is your repro (pack with `adapter_revision="NULL"`, `ALTER TABLE ... UPDATE adapter_revision = NULL ... mutations_sync=2` on the test catalog, Python refuses, native must refuse). The live test was not run locally.

native/csrc/sink/record_row.cpp
zaoxing: Fixed in 96c01e4.

Both copies: `ParseMetadataJson` in `record_row.cpp` takes the empty array before the split (zero dimensions, one element) and still refuses a missing or non-array `shape`; `shape_product` in `hydration.cpp` does the same so `summarize_core` accepts the scalar. Your exact repro (`meta["shape"] = []`, `_row(..., shape=())`, `torch.tensor(3.5)`) is `test_a_scalar_envelope_is_admitted` in `test_native_adapter_torch.py`; the driver path is `test_a_scalar_shape_is_admitted_as_one_element` plus `test_a_missing_shape_is_still_refused` in `test_native_pack_sink.py`; the summary is `test_a_scalar_capture_hydrates_and_summarises` in `test_native_reader_parity_live.py` (element_count 1, mean 3.5 against Python's summarize; live, not run locally). The CPU tests pass here.

native/csrc/sink/record_row.cpp
zaoxing: Fixed in 96c01e4.

`FindIntChecked` now requires the whole token to be the integer: after the digits only a JSON delimiter (`,` `}` `]` whitespace or end) may follow, so `0.5` and `123.5` are refused as not-an-integer rather than scanned as 0 and 123. A new `IntDomain::kUnsigned` refuses a negative literal while the sign is still in the text, and the six unsigned counters use it, so `step_number=-1` is refused instead of wrapping to 18446744073709551615. The shape loop requires each item to be digits only (`[16.5]` is refused).

Tests: `test_invalid_numeric_metadata_is_refused_not_converted` (your table: layer_number 0.5, captured_at_ns 123.5, step_number/producer_rank/token_start/batch_position -1) and `test_a_fractional_shape_dimension_is_refused_not_truncated` in both `test_native_pack_sink.py` and `test_native_adapter_torch.py`, each asserting `CaptureMetadata.from_mapping` refuses the same input and that nothing was submitted. Passing here.

native/csrc/common/json.cpp
zaoxing: Fixed in 96c01e4.

`dmi_common::Unescape` combines a high surrogate followed by `\u` + low surrogate into one code point and encodes it as four UTF-8 bytes; a lone surrogate keeps the previous behaviour. This is the one decoder the sink parser and the pack-index footer reader share, so both paths are fixed together.

Tests: `test_a_non_bmp_identifier_survives_the_metadata_decoder` in `test_native_pack_sink.py` (hook_name `block` + U+1F600 through `submit_row`, read back with Python `PackReader`, bytes compared; `block中文` as the BMP control) and `test_a_non_bmp_hook_name_round_trips` in `test_native_adapter_torch.py`, both passing here. The index-path repro (`SELECT hex(hook_name)` equals `626C6F636BF09F9880`, Python hydrate and native search see the name) is `test_a_non_bmp_hook_name_indexes_intact` in `test_native_reader_parity_live.py`, not run locally.

native/csrc/catalog/hydration.cpp
zaoxing: Correction to my reply above, and the fix is now in a1daa6e.

I had made only the FOOTER side of the comparison null-aware and claimed the catalog side already carried a NULL as the empty field. That was wrong: the reader returns the literal text `NULL` for a SQL NULL inside the argMax tuple (its own parity helper documented this, and the previous `normalize_footer_field` was mapping `"NULL"` to `""` on BOTH sides, which is why the ambiguity you found existed at all). So the first commit refused every capture with `adapter_revision=None` on field 10, and the string-vs-NULL case still bound. That is what the failing clickhouse-live run showed.

The distinction only exists while the quotes are still in the text, so it is now kept in `parse_tsv_tuple`: an unquoted `NULL` token becomes the empty string, a quoted `'NULL'` stays the four characters. Empty is unambiguous here because `adapter_revision` is the schema's only Nullable column and both validators require its text to be non-empty. This also puts the two readers on one representation, since Python already maps that column to `None`, so the parity helper no longer tolerates `"NULL"` for it.

`parse_tsv_tuple` is declared in `reader.h` and reachable through a new session-less `tuple_fields` op, so both halves of the comparison are pinned on the CPU gate in `tests/test_native_hydration_footer_fields.py` (all four NULL/'NULL' combinations, catalog side and footer side, compared the way hydrate compares them). It was live-only before, which is precisely why I shipped a wrong assumption; that test now fails without this fix.

Samfisheryu (approved):
Re-reviewed at 878a6a4; no remaining blocking findings in this review.

Pushed two small, reproduced spool fixes: upload polling no longer deletes an active writer’s .open file, and repeated removal no longer releases another pack’s capacity. Both regression tests failed before the fixes and pass now. Crash cleanup remains explicit, with other writers stopped.

Validation: 1,462 CPU tests, 136 live-ClickHouse tests, and 16 single-GPU E2E tests passed. The GPU checks include 288 CUDA Graph replays with byte-for-byte tensor readback and the legacy reference/ClickHouse paths. No GPU Ring or legacy ClickHouse implementation changes.

S3 validation used the signature-checking fake, not live Garage. Hosted CI for this commit is still running.

PR #129, opened 2026-09-06
feat(catalog): Phase C — native reader at parity with the Python CaptureReader

zaoxing:
First slice of Phase C (serving path), stacked on #128. Checkpoint B passed; this starts C1.

## What

`native/csrc/catalog/reader.h/.cpp` — the read surface of `clickhouse_reader.py`, ported:

- **search**: the projection's one-aggregate shape (argMax over a tuple, ordered by the total `(index_version, store_id, pack_id)`), the bounded membership predicate, equality/IN/range filters, keyset pagination over the 5-column sort key, limit+1 cursor logic.
- **get_by_ids**: tenant-leading (primary index + bloom filter), chunked inline ids, the watermark bound on a deciding read.
- **current_watermark**.

The cursor codec is byte-compatible with `cursor.py`: same JSON envelope (sorted keys, compact separators), unpadded url-safe base64 with canonical-alphabet validation, and the same `filter_hash` (sha256 over the filter dict in Python's exact serialization) — so cursors cross implementations.

## Read-parity suite

`tests/test_native_reader_parity_live.py` (manual+clickhouse): both readers run the same queries against one catalog and return identical 32-field descriptors:

1. search parity, no filters
2. search parity with filters (hook IN, time bounds, tenant)
3. **cursor parity across implementations** — the union of alternating pages covers the corpus in order
4. get_by_ids parity + watermark validation refused by both
5. supersession — a later pack's locator wins on both sides

5/5 × 3 consecutive runs; the native catalog suites stay at 44/44; the CPU gate at 1202 passed.

## Caught by the parity suite before landing

- the aggregate tuple carried the sort-key columns (the mixed-row shape the one-aggregate design forbids),
- the pagination cursor keyed on the row beyond the page instead of the last kept item,
- the base64url encoder emitted a phantom trailing byte in both remainder branches (byte-incompatible with Python's decoder).

## Follow-ups

C2 (hydration/summary/extensions native), C3 (default-switch + final docs).


zaoxing:
**C2 landed (fbdb607) — hydration and core summary at exact parity.**

The hydration orchestration is now native: select → estimate → hydrate (per-pack plans with coalesced ranges, the pack footer verified as the authority before any payload fetch, byte/request budgets refused up front, per-record CRC verify) and `summarize_core` (float64 accumulators, scale-before-square L2, raw-integer order stats). The extension registry stays Python-side by scope — extensions are Python pluggables.

Parity is exact against the Python `CaptureReader` over the full e2e (native sink → uploader → index → select → hydrate through the fake S3): **identical payload bytes**, byte-compatible selection identity, core numbers round-tripping at %.17g, budget refusals agreeing. The suite caught four native bugs before landing: the selection identity hashed in the wrong key order, the footer comparison mixing the two row layouts, the payload CRC compared as decimal against a hex locator, and the summary emission truncating doubles to six places.

9/9 parity × 3 runs; 44/44 native catalog tests; CPU gate 1202 passed. Remaining in Phase C: **C3 — the default-switch (D4) + final docs**.

zaoxing:
**C3 landed (24097b4) — the default-switch. Phase C complete; the plan's end state is reached.**

With `storage_backend="capture"` and no explicit `record_sink`, the engine now builds the **native pack writer** from the config's `capture_sink_config`. An explicit `record_sink` overrides it — the reference sink is the documented rollback, and the rollback guarantee (packs staged by either writer readable by the same Python reader) stays pinned. The ClickHouse host record path (`"native"`) and the `"auto"` inference are untouched: every pre-existing caller sees no change. Without a `capture_sink_config` the refusal remains, now naming the missing field.

**Gates the flip rode on** (D4's condition): Checkpoint A ✓, Checkpoint B (human review of #127/#128) ✓, quorum verifier vs the C++ writer ✓, C1 read-parity ✓, C2 hydration/summary parity ✓.

Docs: the config's capture-backend comment, the sink module's selection docstring, and a ledger entry in `docs/benchmarks.md` that restates the standing obligation — **re-measure on a quiet host before publishing any throughput claim** (this host's numbers were shared).

Engine suite 22/22 (new default test + updated refusal); CPU gate **1203 passed**. The pipeline is complete: Phase 0 → A → B → C, with the native path as the production writer and the Python path as the conformance oracle and documented rollback.

zaoxing:
Copilot's filter_hash thread — **reproduced with two tests** and **fixed across 74784c6/925de3d** (with the parallel review pass): the filter hash AND the cursor's key array rendered values by raw concatenation — a quote/backslash/non-ASCII id broke the cursor JSON outright and hashed different bytes on each side. Both now go through the canonical `EncodeJsonString` (the pack footer escaper, conformance-proven byte-identical to `json.dumps ensure_ascii`), with key values unescaped from raw TSV before rendering. Cursors cross implementations over quote-bearing, backslashed, non-ASCII values; 15/15 parity × 3; CPU gate 1203. Thread resolved.

zaoxing:
**Dtype generality pass landed (a5e15a2) — fp8 and the wide integers are first-class.**

Your compatibility concern, audited and addressed. The dtype tables had drifted: the C++ hydration decoder accepted fp8 and uint16/uint32, but the Python validation layer (`_DTYPE_BYTES`) rejected them at `CaptureMetadata` — the pipeline was only as wide as its narrowest layer. One authoritative 14-dtype table now drives every consumer (numpy summary map, C++ hydration decoder, pack builder, sink), fp8 decodes by exact bit math on both sides (OCP e4m3fn with no infinities, IEEE-shaped e5m2 — NaN/Inf counting included in the summary parity), and the golden corpus covers all 14 dtypes with the manifest regenerated and its test de-hardcoded. New boundary tests pin rank/dim/counter ceilings and the footer capacity bound (loud at write). A "format limits and portability" section in `docs/capture-storage-design.md` states the versioning gates, dtype contract, ceilings, little-endian definition, and checksum layers.

Verification: parity 15/15 × 3, native catalog 44/44, golden/conformance/summary 97 passed, CPU gate 1203.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+                continue;
+            }
+            if (c == '\'') {
+                in_str = true;
+                continue;
+            }
+            if (c == ',') {
+                fields.push_back(current);
+                current.clear();
+                continue;
```
Samfisheryu: [P1] Preserve array nesting when parsing footer rows

This parser splits at every unquoted comma, including commas inside `shape`. A valid `[2,8]` becomes two fields, shifting the remaining footer columns.

Reproduced with float32 payloads shaped `(2,8)` and `(2,2,4)` through native sink → upload → index → select → hydrate. All stages through search succeed; Python hydration returns the original bytes exactly. Native hydration fails:

```text
catalog descriptor does not match the pack footer: field 20 for upload-0001
```

The search-side parser in `reader.cpp` already handles nested arrays, but this separate parser does not. Please share a correct parser (or keep typed descriptors instead of round-tripping through SQL text) and add multidimensional tensors to the hydration E2E, not only the search parity test.

src/dmi/storage/capture/model.py
```diff
@@ -8,14 +8,22 @@
...
 _DTYPE_BYTES = {
     "bool": 1,
     "uint8": 1,
     "int8": 1,
+    "uint16": 2,
     "int16": 2,
     "float16": 2,
     "bfloat16": 2,
+    "float8_e4m3fn": 1,
+    "float8_e5m2": 1,
```
Samfisheryu, line 24: [P2] Carry these new dtypes through the native Ring adapter too

The four new names are admitted here, but `AtenDtypeName` in `native/csrc/sink/native_pack_sink.cpp` still handles only the old ten types. All four fail at the adapter boundary.

Reproduced using `_make_sink`/`_meta` from `test_native_adapter_torch.py`: submit a valid 16-element row with the matching byte length and ATen code:

```text
float8_e4m3fn: 24, float8_e5m2: 23, uint16: 27, uint32: 28
All four: RuntimeError: NativePackSink: unsupported payload dtype
```

The metadata and pack-layer acceptance therefore does not make these dtypes usable through the actual capture adapter. Please extend the adapter mapping and parameterize its torch-envelope tests over the new types, alongside the format/summary tests.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+    // pattern, matching summary.py's _decode_float8 bit for bit.
+    // e4m3fn: 1/4/3, bias 7, no infinities (all-ones exponent is NaN).
+    // e5m2: IEEE-shaped 1/5/2, bias 15, with infinities.
+    const uint8_t byte = data[offset];
+    double sign = byte >= 128 ? -1.0 : 1.0;
+    const uint32_t body = byte & 0x7F;
+    if (dtype == "float8_e4m3fn") {
+      const uint32_t exponent = body >> 3;
+      const double mantissa = static_cast<double>(body & 0x7);
+      if (exponent == 15) return std::nan("");
```
Samfisheryu: [P1] Preserve the finite exponent-15 values in e4m3fn

For e4m3fn, exponent 15 is not entirely NaN: only the mantissa-7 encoding is NaN. For example, `0x7e` is **448**, but this branch returns NaN. The same error is present in Python `_decode_float8`, so the current parity test agrees with an incorrect oracle (it also labels `0x7e` as NaN).

Independent repro:

```python
import numpy as np
import torch
from dmi.storage.capture.summary import _decode_float8
raw = bytes(range(256))
expected = torch.arange(256, dtype=torch.uint8).view(torch.float8_e4m3fn).float().numpy()
actual = _decode_float8(np, raw, 'float8_e4m3fn')
# 14 mismatches: 0x78..0x7e and 0xf8..0xfe.
# Finite values +/-256..448 become NaN.
```

Please fix both decoders and use the independent conversion as the test oracle. This changes decoded tensor/statistical values, even though the raw pack payload bytes remain intact.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+      }
+      finite.push_back(value);
+      sum += value;
+      if (value == 0.0) ++zeros;
+      if (first || value < min_f) min_f = value;
+      if (first || value > max_f) max_f = value;
+      scale = std::max(scale, std::fabs(value));
+      if (summary.order_stats_are_integers) {
+        const int64_t as_int = static_cast<int64_t>(value);
+        if (first || as_int < min_i) min_i = as_int;
```
Samfisheryu, line 820: [P2] Compute integer extrema before converting values to double

`decode_element` first converts int64 to double, and this cast converts that rounded double back to int64. The resulting minimum/maximum are not exact integers above 2**53, contrary to the summary contract.

Repro: stage and index a one-element `int64` tensor with payload `struct.pack('<q', 2**53 + 1)`, then call `summarize_core`:

```text
expected minimum_int: 9007199254740993
actual minimum_int:   9007199254740992
```

Please compute integer order statistics directly from the raw integer dtype; reserve float64 conversion for the accumulation fields where rounding is allowed. Include values above 2**53 and signed endpoints in the regression tests.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+            std::to_string(max_summary_captures));
+  }
+  const auto payloads = hydrate(selection, byte_limit, request_limit);
+  const auto descriptors = resolve(selection);
+  uint64_t total_elements = 0;
+  for (size_t i = 0; i < descriptors.size(); ++i) {
+    total_elements +=
+        shape_product(descriptors[i][kShape]) *
+        (parse_u64_field(descriptors[i][kDecodedLength], "decoded") /
+         dtype_bytes(descriptors[i][kDtype]));
```
Samfisheryu: [P1] Count each tensor's elements once, not N squared

Both `shape_product(...)` and `decoded_length / dtype_bytes` already equal the number of elements. Multiplying them turns a normal N-element tensor into N² for the limit check.

Repro: write/upload/index one `uint8` tensor with shape `(9000,)` and 9000 payload bytes, then use the native driver's default `summarize_core` limits:

```text
summary element limit exceeded: 81000000 > 64000000
```

There are only 9000 elements, so this should easily fit. Please use one element count, validate the shape/byte consistency separately, and check the summary budget before fetching payloads.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+      }
+      // Metadata and locator must agree. The two layouts name fields at
+      // different indexes: the catalog row is sort-key-first (capture_id
+      // 4, dtype 19, shape 20); the footer row is CAPTURE_COLUMNS order
+      // (capture_id 0, dtype 18, shape 19). Everything else lines up.
+      static constexpr std::pair<size_t, size_t> kCompared[] = {
+          {kCaptureId, 0},   {kDtype, 18},       {kShape, 19},
+          {kObjectKey, 23},  {kObjectBytes, 24}, {kPayloadOffset, 27},
+          {kStoredLength, 28}, {kDecodedLength, 29},
+          {kPayloadChecksum, 31}};
```
Samfisheryu: [P2] Compare the complete descriptor metadata with the authoritative footer

This comparison omits identity metadata such as hook name, layer, request, and tenant. Consequently it can accept a payload attached to metadata that disagrees with the pack, while the Python reader rejects it.

Repro on an isolated test catalog: write/upload/index a valid uint8 `(16,)` record, then change only the catalog row's `hook_name` to `wrong-hook` using a synchronous test-table mutation. Python `CaptureReader.hydrate` refuses the footer mismatch; native `hydrate` returns `ok=true` and the 16 payload bytes.

This is fault injection against the existing 'footer is authoritative' contract, not a claim that normal single-stream capture reorders metadata. Please compare all authoritative metadata and locator fields before serving payloads, matching the Python reader.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+  for (const auto& plan : plans) {
+    const auto& first_descriptor = descriptors[plan.descriptor_indexes[0]];
+    PackRefData ref{
+        plan.pack_id, plan.store_id, plan.object_key,
+        parse_u64_field(first_descriptor[kObjectBytes], "bytes"),
+        first_descriptor[kPackChecksum],
+        parse_u64_field(first_descriptor[kPackRecordCount], "records")};
+    // The footer's own descriptor rows, one per record: the same 32-field
+    // renderer the indexer reads, with the batch's version irrelevant here.
+    const auto footer_rows = read_pack_descriptor_rows(s3_, ref);
```
Samfisheryu, line 584: [P1] Charge trailer and footer reads against the hydration budget

The preceding guards count payload ranges only. This call then performs two extra S3 reads without consuming/checking the remaining request and byte budgets.

Repro with one uncached 16-byte uint8 capture: call `hydrate` with `byte_limit=16, request_limit=1`. Python refuses because the payload already uses the entire allowance. Native returns `ok=true` after **three GETs** (trailer, footer, payload), exceeding both limits.

Please pass a shared remaining-budget object through footer reads and payload reads, refusing each fetch before issuing it. Including footer cost in `estimate()` is not sufficient if `hydrate()` bypasses the configured limits.

native/csrc/catalog/hydration.cpp
```diff
@@ -0,0 +1,707 @@
...
+        payloads[index] = payload;
+      }
+    }
+  }
+  for (const auto& payload : payloads) {
+    if (payload.empty()) {
+      // (a zero-length payload is legal only if the locator said so; our
+      //  v1 packs never produce one, and resolve() guarantees every slot)
+      throw CatalogError(CatalogError::Kind::kValue,
+                         "hydration plan did not resolve every capture");
```
Samfisheryu: [P2] Distinguish a resolved empty tensor from an unresolved capture

A zero-byte payload is legal when a dimension is zero; it cannot be used as the sentinel for an unfilled result slot.

Reproduced with `dtype='float32', shape=(0,), payload=b''`: native pack → upload → index succeeds, and Python hydration returns the empty bytes correctly. Native hydration reaches this final check and fails with:

```text
hydration plan did not resolve every capture
```

Please track whether each slot was resolved separately from its payload length. Add a zero-element tensor E2E so legal empty captures are not rejected.

Samfisheryu (commented):
Eight reproduced correctness issues are inline. I would hold the native-default switch until these and the Ring-binding/spool issues reported on #127 are fixed. In particular, normal multidimensional tensors cannot hydrate, and finite FP8 values are decoded as NaN on both the native and Python sides.

Tests at this head: 78 existing native pack/store/sink tests, 66 real-ClickHouse catalog/reader tests, and 10 existing single-GPU reference/ClickHouse Ring tests passed (these suites overlap with broader CPU coverage). The new default capture runtime itself fails at extension loading; that is not a successful new-path GPU E2E. Independent boundary probes produced 17 failing cases out of 18, grouped into the 14 findings across this stack. S3 was a local signature-verifying fake, not live Garage; no HF/vLLM graph, TP, or production stress-test coverage is claimed.

The FP8 result is especially important: comparing the C++ decoder against the Python decoder cannot catch a mistake copied into both. The independent oracle here was PyTorch conversion across all 256 FP8 byte patterns.

native/csrc/catalog/hydration.cpp
zaoxing: Reproduced with a test (test_zero_element_tensor_hydrates_to_empty_bytes — a float32 (0,) tensor through sink→upload→index→select→hydrate; Python returned the empty bytes, the native reader refused) and fixed in 79addbb: the resolution sentinel is a separate flags vector, decoupled from the payload length. A zero-element tensor now hydrates to empty bytes on both sides.

PR #131, opened 2026-09-07
[Core] Add recurring D2H window scheduling

XbzOnGit:
## Purpose

Add an opt-in D2H policy for frameworks that know recurring execution regions where offload traffic is less likely to contend with TX-critical work. The framework supplies only a repeating window pattern and ordered boundaries; DMI-core learns how many complete-record bytes each window can carry.

The existing GPU ring, drain stream, record reconstruction, P2P thread, and sink continue to move and publish the data. The feature is disabled by default and applies only to the HookPointV1 record path.

## Configuration

```python
window = dmi.RecurringD2HWindowConfig()
window.enabled = True
window.minimum_record_probe_retry_interval_occurrences = 4
window.timing_revalidation_retry_interval_occurrences = 4
window.capacity_flush_fallback_threshold = 3
window.capacity_flush_count_reset_interval_periods = 32

ring = dmi.RingConfig()
ring.recurring_d2h_windows = window

engine = dmi.MonitoringEngine(
    model_id="my-run",
    host_engine=host,
    ring_config=ring,
    record_mode_v1=True,
)
runtime = engine.create_record_runtime(record_format)
```

The integration defines sorted, non-overlapping half-open windows within a positive period `T`:

```python
runtime.define_d2h_window_pattern(
    period=T,
    windows=[(begin_0, end_0), (begin_1, end_1)],
    initial_counter=0,
)
```

Window `(begin, end)` in repetition `n` is `[n*T + begin, n*T + end)`. The integration then publishes each ordered boundary on its framework CUDA stream:

```python
runtime.advance_boundary()
```

`enabled` defaults to `False`. Enabled configurations require positive probe intervals and fallback threshold. Pattern definition returns `False` only after terminal fallback rejects further definitions. Debug mode logs real definitions, transfers, outcomes, warnings, and fallback.

Megatron exposes the feature through `--dmi-recurring-d2h-windows`. It supports eager, non-interleaved 1F1B with PP>1 and the standard pipeline communicator. It requires `batch_p2p_sync=False`; when both synchronous P2P and windowing are requested, the integration prints a warning and disables windowing. PP=1 and interleaved/VPP execution retain ordinary batching; multi-module pipelines are rejected when the option is enabled.

`batch_p2p_sync` is a workaround for incorrect asynchronous batched-P2P ordering in older PyTorch releases. The coalesced NCCL work/stream-ordering fix landed in PyTorch 2.1 through [pytorch/pytorch#98793](https://github.com/pytorch/pytorch/pull/98793). This Megatron checkout already requires `torch>=2.6.0`, and the validation below used PyTorch `2.13.0+cu129`. An older affected stack that must keep `batch_p2p_sync=True` therefore retains ordinary DMI batching instead of enabling recurring windows.

## Algorithm

Each window position learns:

- `A`: largest proven-safe transfer;
- `B`: smallest proven-unsafe transfer; and
- `E`: clean same-window timing estimate.

The `BINARY_ADAPTIVE` policy starts with all currently transferable records. With only `B`, it halves; with both bounds, it uses their midpoint; with only `A`, it uses `A` or an allowed larger `E`. Completion before CLOSE raises `A`; completion at or after CLOSE lowers `B`.

If a transfer at or below `A` overruns, the controller treats it as a downward capacity shift, clears stale safe/timing evidence, and resumes halving. Timing evidence is used only when OPEN observation, issue, completion, and CLOSE observation belong to the same occurrence.

Records are never split. A first record larger than the target may be probed as one unit. Repeated failed minimum-record and timing-revalidation probes use increasing occurrence cooldowns, preventing a failed probe every recurrence. Indivisible-record convergence stalls at the current safe prefix and emits one warning; record chunking is outside this PR.

An overrun crossing later boundaries updates only its originating window and resumes from current progress. Repeated eligible capacity flushes trigger terminal fallback to ordinary batching; isolated evidence ages out after the configured number of periods. Lifecycle, shutdown, and oversized-record flushes do not count.

## Implementation details

```text
framework pattern + boundaries
  -> GPU-authoritative version/counter
  -> pattern matcher + adaptive controller
  -> DrainThread admission
  -> existing D2H/reconstruction/P2P/sink path
```

`advance_boundary()` enqueues a captureable GPU operation on the framework stream. It increments a packed 16-bit version / 48-bit counter and publishes it to CPU-visible memory without a host callback or GPU/CPU synchronization.

Forced flush handling remains first in the drain loop. In an active window, the controller admits at most one transfer per occurrence. The drain packs the largest FIFO prefix of whole ready records that fits, rechecks the version and occurrence, issues D2H on the existing copy stream, and reports actual bytes and the completion boundary. Disabled, no-pattern, missed-window, and fallback states proceed to the existing `should_flush()` batching path.

Pattern definitions are versioned and pending definitions remain ordered. Completion from an older version cannot update current learning state. At exceptional version exhaustion, DMI synchronizes the framework stream, flushes and pauses the drain worker, clears version state, enqueues the reset and new definition, then resumes.

## Compatibility and scope

The current Megatron path continues to pass its native `DMXHostEngine` through `storage_backend="auto"` and creates the record runtime without an explicit `record_sink`. It therefore retains the ClickHouse tensor-bytes path rather than selecting the capture/object-storage sink.

This PR does not infer windows inside DMI-core, coordinate learning across ranks/streams/engines, split records, or claim that a logical window proves physical PCIe/NIC idleness.

Pinned integration revisions:

- DMI-Megatron-Integration: [`f416d73`](https://github.com/ProjectDMX/DMI-Megatron-Integration/commit/f416d73d9e1a4e93b7ff97e83c83d00041b57b08)
- Megatron-LM-DMI: [`1aecebd`](https://github.com/ProjectDMX/Megatron-LM-DMI/commit/1aecebd768834855349ac4f1b5b8d2003f277a16)

## Validation

- Native grant controller: 203 passed.
- Native ring engine: 147 passed.
- Python window bindings: 5 passed.
- DMI runtime/storage and window API against latest `main`: 28 passed.
- Megatron startup, schedule, boundary, and trace analysis: 147 passed.
- Native CUDA rebuild/link and `git diff --check`: passed.

A 13-iteration Qwen3-1.7B-shaped PP=2 run with `batch_p2p_sync=False` compared ordinary batching and recurring windows while offloading forward hidden states. Both modes stored exactly 728 logical records and 3,053,453,312 payload bytes, with zero bidirectional identity differences.

For placement analysis, we conservatively treat compute windows as TX-critical. In large-scale training, these regions may also carry cross-node distributed traffic such as EP communication, so D2H traffic should avoid them.

| Mode | Large D2H | Issued in Preferred Window | Completed in Preferred Window | D2H in TX-critical windows |
|---|---:|---:|---:|---:|
| Ordinary batching | 2,912 MiB | N/A | N/A | 71.676 ms |
| Recurring windows | 2,912 MiB | 100.00% | 96.58% | 2.686 ms |

Trace-measured D2H overlap with TX-critical windows fell by 96.25%. The controller recorded 34 successes and four overruns without minimum-record probes or terminal fallback. The short window converged to a 4 MiB grant after its larger grants overran; the long window remained successful.

Under Nsight, iterations 3--13 averaged 479.791 ms with ordinary batching and 470.591 ms with recurring windows (-1.92%) in this single run. This validates traffic placement and downward reconvergence.


native/csrc/ring/recurring_d2h_grant_controller.cpp
```diff
@@ -0,0 +1,376 @@
...
+            overran ? D2HWindowCompletionResult::OVERRUN
+                    : D2HWindowCompletionResult::SUCCESS);
+    }
+    if (actual_bytes == 0 || !current_bundle_ ||
+        current_bundle_->version != admission.version) {
+        return;
+    }
+
+    auto& state = current_bundle_->windows.at(admission.window.window_index);
+    const auto policy_observation = state.policy->observe(D2HWindowAttempt{
```
zaoxing, line 377: **[correctness] Late-issue overrun wipes learned `max_safe_` as a capacity shift**

`complete()` feeds every overrun to `policy->observe()`, which treats `overran && bytes <= prior_max_safe` as a downward capacity shift and resets `max_safe_`/`timing_estimate_` — even when the transfer only overran because the drain first polled this occurrence late (after a forced flush, a long `sync_stream`, or a cross-window overrun that already set `missed_open_occurrence`). The `clean_open`/`missed_open` knowledge only gates timing evidence, not the capacity observation.

Scenario: W has converged to A=64 MiB. `reserve_record()` forces a 30 ms flush ending 1 ms before W's CLOSE; the next poll lands inside W with `spent_occurrence` unset, admits MAX_SAFE 64 MiB at CLOSE−1 ms, and it overruns → `max_safe_.reset(); min_unsafe_ = 64 MiB`. The next occurrences re-converge from 32 MiB with under-filled windows, and this repeats after every drain stall. (The validation note about the short window converging to 4 MiB after larger grants overran is consistent with this.)

Suggested fix: in `poll()`, skip (or flag in the decision) an occurrence whose `timing->clean_open` is false / `missed_open_occurrence` matches, and have `observe()` not reset `max_safe_` for such attempts.

src/dmi/engine.py
```diff
@@ -254,7 +261,7 @@ def create_record_runtime(
 
         transport = self._ring_transport
         ring_config = self._ring_config
-        if transport is None or ring_config is None:
+        if ring_config is None:
```
zaoxing, line 264: **[correctness] `create_record_runtime()` now succeeds after `close()`**

The guard was relaxed from `transport is None or ring_config is None` to `ring_config is None`, but `close()` clears `_ring_transport`/`_ring_engine`/`_record_mode`/`_host_engine` and never clears `_ring_config`. So `engine.close(); engine.create_record_runtime(fmt)` — which previously raised `RuntimeError('Ring transport is not enabled')` — now skips host-schema validation (`_host_engine` is None), calls `RingEngine.create_record(ring_config, None)` (a drain/P2P worker with no sink, records silently dropped), and `_rt.activate()`s the transport on an engine the caller believes is torn down.

Suggested fix: clear `self._ring_config` in `close()` (or track a closed flag and raise).

native/csrc/ring/recurring_d2h_grant_controller.cpp
```diff
@@ -0,0 +1,376 @@
...
+    const D2HWindowProgressSnapshot& progress, D2HWindowClock::time_point now) {
+    if (!current_bundle_ || progress.version != current_bundle_->version)
+        return;
+
+    for (auto& state : current_bundle_->windows) {
+        if (!state.timing.has_value() ||
+            progress.counter < state.timing->absolute_close) {
+            continue;
+        }
+        const auto estimate = estimate_window_bytes(*state.timing, now);
```
zaoxing: **[correctness] CLOSE is observed at the first poll after close, inflating the timing estimate**

`observe_progress()` closes a `TimingObservation` with `now` at the first poll where `counter >= absolute_close`. Any drain stall between the real CLOSE and that poll (forced full flush from `reserve_record`/`prepare_step`, or a long `sync_stream` in another window) inflates `window_ns` and thus the byte estimate. The clean check (`open <= issue < completion < close`) cannot detect this, and in the only-A branch `choose_base()` grants the estimate with no revalidation gating.

Scenario: A=32 MiB, a 32 MiB transfer completes 5 ms into a 10 ms window; a 200 ms forced flush delays the next poll → estimate ≈ 32 MiB × 205/5 ≈ 1.3 GiB → next occurrence issues a 1.3 GiB TIMING_ESTIMATE grant → guaranteed overrun into the TX-critical region.

Suggested fix: accept the estimate only when the close was observed promptly, e.g. require `progress.counter == state.timing->absolute_close` (not beyond), or bound `now - open` by the expected window duration / poll timeout.

native/csrc/ring/drain_thread.cpp
```diff
@@ -162,6 +179,38 @@ bool DrainThread::force_flush_and_wait_until(
...
+    });
+    if (drain_failure_) std::rethrow_exception(drain_failure_);
+    return DrainPauseToken{generation};
+}
+
+void DrainThread::resume(DrainPauseToken token) {
+    {
+        std::lock_guard<std::mutex> lock(mu_);
+        if (token.generation == 0 ||
+            token.generation != pause_acknowledged_generation_ ||
```
zaoxing, line 223: **[correctness, latent] `resume()` silently drops a mismatched token → drain stays paused**

`resume()` is a no-op when `token.generation != pause_acknowledged_generation_`. The loop acknowledges only the latest `pause_requested_generation_`, but every earlier waiter's predicate (`acknowledged >= generation`) also fires, so two concurrent `pause_after_flush_and_wait()` callers each get a token yet only the newest can ever resume; the loop's `cv_.wait` on `pause_resumed_generation_ >= pause_generation` then blocks forever and every later `force_flush_and_wait()`/`reserve_record()` hangs.

Today only the subsystem calls this and `version_reuse_in_progress_` serialises it, but `DrainPauseControl` is a generic interface and the silent return hides the hang. Consider treating any token `<= acknowledged` as resuming (track an outstanding count) or throwing on mismatch.

native/csrc/ring/drain_thread.cpp
```diff
@@ -370,6 +421,85 @@ void DrainThread::do_full_flush() {
...
+
+        uint64_t staging_free = 0;
+        {
+            std::lock_guard<std::mutex> staging_lock(staging_mu_);
+            staging_free = staging_.free_bytes();
+        }
+
+        uint64_t full_count = 0;
+        uint64_t full_bytes = 0;
+        for (const uint64_t actual_bytes : scanned_) {
```
zaoxing: **[efficiency] `do_window_decision` scans all pending entries every poll**

This walks the entire `scanned_` deque under `mgmt_mu_` on every drain iteration, although `pending_bytes_` already holds the aligned sum and the scan is only needed when `pending_bytes_ > staging_free`. In ACTIVE mode data is deliberately held back, so `scanned_` grows toward `task_ring_entries` (65536 default) and this O(pending) loop repeats every `drain_poll_timeout_us` (100 µs default) while holding the lock `prepare_step()`/`reserve_record()` need.

Cheaper: if `pending_bytes_ <= staging_free`, use `full_count = pending_entries_`, `full_bytes = pending_bytes_`; otherwise scan (or cache the prefix between polls, since `scanned_` only grows at the back).

native/csrc/ring/d2h_window_grant_policy.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+D2HWindowGrantDecision BinaryAdaptiveGrantPolicy::choose_base(
+    uint64_t occurrence, D2HWindowAvailability availability) const {
+    D2HWindowGrantDecision decision;
+    decision.full_grant_bytes = availability.full_grant_bytes;
+    decision.prior_max_safe = max_safe_;
+    decision.prior_min_unsafe = min_unsafe_;
+
+    if (!max_safe_.has_value() && !min_unsafe_.has_value()) {
+        decision.byte_limit = std::numeric_limits<uint64_t>::max();
```
zaoxing, line 74: **[altitude] First grant is unbounded, and an overrun discards its timing evidence**

With no bounds the policy grants `UINT64_MAX` (everything that fits staging); when that overruns, `complete()` resets the timing observation, so the controller learns only B=huge and halves from there. From a multi-GiB backlog that is ~log2(backlog/capacity) consecutive overruns into the TX-critical regions the feature exists to protect (4 GiB → 2 → 1 → 512 → … → 4 MiB ≈ 10 occurrences).

The overrunning transfer's completion time already gives a first-order estimate (`bytes × window_ns / transfer_ns`) that is thrown away. Consider capping the initial grant (configured seed or `staging_free / k`) and/or using the overrun's timing as the initial B.

native/csrc/ring/recurring_d2h_grant_controller.cpp
```diff
@@ -0,0 +1,376 @@
...
+        ++matching_index;
+    }
+    if (matching_index < pending_bundles_.size()) {
+        for (size_t index = 0; index < matching_index; ++index)
+            pending_bundles_.pop_front();
+        current_bundle_ = std::move(pending_bundles_.front());
+        pending_bundles_.pop_front();
+        cached_progress_ = observed;
+        mode_.record_pattern_version_activation();
+        std::fprintf(stderr, "[d2h_window] active version=%u\n",
```
zaoxing: **[nit] Unconditional stderr log on every version activation**

This `[d2h_window] active version=%u` print is not gated on `debug_enabled`, unlike the other informational window logs that go through `D2HWindowDebugLogger`. An integration that redefines the pattern per phase will spam every rank's stderr in production. The terminal-fallback message is a legitimate warning; this one belongs behind `debug_logger_`.

native/csrc/ring/d2h_window_subsystem.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include <cstdio>
+#include <stdexcept>
+#include <string>
+#include <utility>
+
+namespace ring {
+namespace {
+
+void check_cuda(cudaError_t error, const char* operation) {
```
zaoxing: **[reuse] `check_cuda` duplicated**

An identical anonymous-namespace `check_cuda` is also defined in `d2h_window_progress.cpp`, `d2h_window_marker.cu` adds a near-identical `check_launch`, and `drain_thread.cpp` already has `throw_cuda_failure` with the same shape. A shared inline helper (e.g. in `d2h_window_progress.h` or a small `ring/cuda_check.h`) would remove three copies and unify the message prefix.

src/dmi/transport/ring.py
```diff
@@ -200,6 +208,31 @@ def __init__(self, ring_engine: Any) -> None:
...
+        self._d2h_window_marker = None
+        self._d2h_window_pattern_defined = False
+        self._d2h_window_device_progress: Optional[torch.Tensor] = None
+        self._d2h_window_cpu_visible_progress: Optional[torch.Tensor] = None
+        window_enabled = getattr(
+            ring_engine, "_recurring_d2h_windows_enabled", None
+        )
+        if window_enabled is not None and bool(window_enabled()):
+            marker_registry = {
+                0: torch.ops.ring.advance_boundary.default,
```
zaoxing: **[nit] Marker registry keyed by literal `0`, not the enum**

This hardcodes the numeric value of `D2HWindowProgressKind.PACKED_VERSION_COUNTER`. If a second kind is added or values reorder, the lookup silently binds the wrong kernel. Keying by `int(D2HWindowProgressKind.PACKED_VERSION_COUNTER)` (or having the engine return the enum) ties it to the source of truth.

docs/config.md
```diff
@@ -77,6 +77,51 @@ By default, timeout-based flushing is disabled and the drain thread flushes at
...
+`record_mode_v1=True`, then create a `RecordRuntime` before defining a pattern.
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `enabled` | `bool` | false | Enable recurring-window scheduling. |
+| `progress` | native enum | packed version/counter | Fixed progress backend for this version; leave unchanged. |
+| `grant_policy` | native enum | binary adaptive | Fixed byte-grant policy for this version; leave unchanged. |
+| `minimum_record_probe_retry_interval_occurrences` | `uint64_t` | 0 | Window occurrences between minimum-record probes; must be positive when enabled. |
+| `timing_revalidation_retry_interval_occurrences` | `uint64_t` | 4 | Base window occurrences between failed timing-revalidation probes; must be positive when enabled. |
+| `capacity_flush_fallback_threshold` | `uint64_t` | 0 | Completed capacity-forced flushes before permanent batched fallback; must be positive when enabled. |
```
zaoxing, line 93: **[docs] Table omits `capacity_flush_count_reset_interval_periods`**

This table lists seven fields but not `capacity_flush_count_reset_interval_periods`, which the PR adds, validates (must be > 0), exposes in the bindings, and documents in `docs/integration-api-v1.md`. The two docs currently describe different field sets for the same struct.

zaoxing (commented):
Review of the recurring D2H window scheduling change. 10 inline comments below — 4 correctness, 1 efficiency, 1 altitude, 2 simplification, 1 reuse, 1 docs.

The two I'd prioritise before merge:

1. **Late-issue overruns wipe learned capacity** (`recurring_d2h_grant_controller.cpp:342`) — the controller already tracks `clean_open`/`missed_open_occurrence` but only uses it to suppress timing evidence; an overrun after a drain stall still resets `max_safe_` as a "capacity shift".
2. **`create_record_runtime()` after `close()`** (`engine.py:264`) — the relaxed guard plus `close()` not clearing `_ring_config` lets a closed engine build a sink-less record ring and re-activate the transport.

Scope note: the `third_party/DMI-Megatron-Integration` submodule bump was not reviewed.

zaoxing:
@XbzOnGit Also a quick question: how was the parameter determined?

window = dmi.RecurringD2HWindowConfig()
window.enabled = True
window.minimum_record_probe_retry_interval_occurrences = 4
window.timing_revalidation_retry_interval_occurrences = 4
window.capacity_flush_fallback_threshold = 3
window.capacity_flush_count_reset_interval_periods = 32

ring = dmi.RingConfig()
ring.recurring_d2h_windows = window

engine = dmi.MonitoringEngine(
    model_id="my-run",
    host_engine=host,
    ring_config=ring,
    record_mode_v1=True,
)
runtime = engine.create_record_runtime(record_format)

src/dmi/engine.py
Samfisheryu: Fixed in a544600. close() now clears the retained Ring config, so neither an initialized engine nor deferred record_mode_v1 construction can be restarted after close. Added both regression cases; they fail before the fix and pass afterward. The complete CPU suite (1,129 tests) and the existing single-GPU reference/ClickHouse tests also pass.

native/csrc/ring/recurring_d2h_grant_controller.cpp
Samfisheryu: Independently reproduced on the current code using RecurringD2HGrantController with an injected clock/progress source (no scheduler sleeps or assumed multi-stream reordering). Period=10, window=[1,3]: first complete 64 MiB cleanly at counter 2; close at the same clock instant to avoid timing evidence. For the next occurrence, first poll at counter 12 (OPEN=11 was missed), issue the learned 64 MiB, and complete at counter 13. At counter 21 the result is: old_safe_MiB=64, safe_retained=0, next_grant_MiB=32. Thus a late issue still erases safe capacity instead of excluding this occurrence from capacity learning. This remains open; a544600 only fixes CUDA API compilation and engine close lifecycle.

native/csrc/ring/recurring_d2h_grant_controller.cpp
Samfisheryu: Independently reproduced with the real controller and a manual clock: period=10, window=[1,3]; issue 64 MiB at counter 1 / t=0, complete at counter 2 / t=5 ms, then next poll at counter 9 / t=205 ms (CLOSE=3 was missed). At counter 11, with a 4 GiB backlog, the next decision reports prior_safe_MiB=64 and next_grant_MiB=2624. The delayed CPU observation is treated as 205 ms of usable window time: 64 * 205 / 5 = 2624 MiB. This confirms the bookkeeping bug without relying on physical-link timing. It is still present after a544600; missed-close observations should not supply a clean timing estimate.

third_party/DMI-Megatron-Integration
```diff
@@ -1 +1 @@
-Subproject commit b0e141f0339095caf67ead3853d909b6265f0d3f
+Subproject commit b9b2c49c2a0e09f3b49332cf5f630ecc4ad40d5f
```
Samfisheryu, line 1: [P2] Suspend training windows during validation, then resume them

The pinned [Megatron schedule](https://github.com/ProjectDMX/Megatron-LM-DMI/blob/19d1999e2057efe24d7372f902bbe177191933f4/megatron/core/pipeline_parallel/schedules.py#L2105) stops publishing boundaries for forward_only, but the [integration runtime](https://github.com/ProjectDMX/DMI-Megatron-Integration/blob/b9b2c49c2a0e09f3b49332cf5f630ecc4ad40d5f/src/dmi_megatron_integration/schedule_runtime.py#L122) leaves the native training pattern active. Validation records then fill the Ring while the last training window is closed. Capacity flushes count against the training policy; after three, it permanently enters ENABLED_FALLBACK, including when training resumes.

Two-part reproduction:

1. Add this test to the pinned integration's tests/test_megatron_d2h_window_schedule.py, using its existing schedule_harness (real schedule, mocked compute/P2P):
```python
def test_train_then_validation_suspends_windows(schedule_harness):
    run, events, runtime, records = schedule_harness
    run(2, 0, 2)
    assert records.counter == 4
    events.clear()
    run(2, 0, 2, forward_only=True)
    assert records.counter == 4
    assert not any(e[0] in ('open', 'close', 'define') for e in events)
    assert not runtime.d2h_windows_active  # fails: still True
```
2. Independently on a real GPU Ring/reference sink: enable windows with period=2, windows=[(1,2)], 64 KiB payload/staging rings, and fallback threshold=3. Publish OPEN, wait for ACTIVE, publish CLOSE, then capture ten 32 KiB tensors without further markers, matching forward-only scheduling. Flush and read back.
```text
evaluation: mode=D2HWindowMode.ENABLED_FALLBACK forced_flushes=3
[d2h_window] terminal fallback count=3 threshold=3
```
All 10 payloads were byte-exact. The problem is that a normal validation phase permanently disables the feature. Please explicitly suspend/resume the native window policy around forward-only phases, using normal batching without charging these flushes as training-window failures. Merely stopping boundary markers is insufficient.

Samfisheryu (changes requested):
Reviewed the DMI changes and the actual pinned Megatron integration at b9b2c49 / Megatron 19d1999, including the opt-in boundary and existing Ring FIFO behavior.

Pushed two small fixes in a544600: the new progress allocation used obsolete CUDA API signatures and broke the CUDA 13 build even with windows disabled; close() also retained the config and allowed a closed/deferred engine to create a record runtime. The latter now has both regression cases.

Requesting changes for the remaining runtime behavior, not documentation: I reproduced Alan's late-issue capacity reset and late-CLOSE estimate inflation in their existing threads. The new inline comment covers validation leaving the training window controller active and causing terminal fallback.

Validation: 1,129 CPU tests and 213 pinned-integration tests passed; native controller tests passed 203 assertions; native Ring tests passed 147 assertions (the multi-device ownership case was skipped with one visible GPU). Ten existing single-GPU reference/ClickHouse tests passed. An additional CUDA Graph test passed 64 replays plus warmup, reading all 65 tensor payloads back byte-for-byte. The failing validation reproduction also preserved all 10 tensor payloads: it breaks window scheduling, not tensor contents. This is not a full multi-GPU Megatron training run.

native/csrc/ring/recurring_d2h_grant_controller.cpp
zaoxing: Fixed in fc5a0e2. `D2HWindowGrantDecision` now carries `clean_occurrence`, and `BinaryAdaptiveGrantPolicy::observe()` skips the entire overrun branch when it is false — no `max_safe_` reset and no `min_unsafe_` tightening. Success still counts, since bytes that fit a shortened window fit a full one.

`clean_open`/`missed_open_occurrence` alone could not see your scenario: after a forced flush the controller has no idea it polled late, so it marks the occurrence clean. The drain now reports its own stalls instead — `note_observation_gap()` before a forced full flush and before an ordinary batching flush — and the first observation after a gap is marked unclean.

`test_late_issue_overrun_keeps_learned_capacity` reproduces @Samfisheryu's counters (period=10, window=[1,3), 64 MiB proven at occurrence 0, OPEN=11 missed, first poll at 12, overrun at 13) and asserts occurrence 2 still grants 64 MiB MAX_SAFE rather than 32 MiB HALF_UNSAFE. `test_truncated_occurrence_overrun_is_not_capacity_evidence` pins the policy directly, including that a *clean* overrun at or below `max_safe_` is still treated as a capacity shift. Both fail before the change.

src/dmi/engine.py
zaoxing: One correction on the state of this thread: **a544600 is not on the PR branch.** It exists in the repo and its parent is dd5627b, but it is the head of no branch, so neither this fix nor the CUDA 13 one was ever reachable from the PR — `close()` at dd5627b still retained `_ring_config`, and this finding still reproduced.

Cherry-picked as c7214d1, so the fix and both regression cases are now actually on the branch.

The CUDA half needed one change, in bb36a5b. It used the `cudaMemLocation` overloads of `cudaMemAdvise`/`cudaMemPrefetchAsync` unconditionally. CUDA 13 does declare only those, but it publishes no `cudaMemAdvise_v2` alias to spell them on older toolkits, and `docs/install.md` still supports CUDA 12.x — so as written it fixed 13 by breaking 12.x. It now selects on `CUDART_VERSION`.

native/csrc/ring/recurring_d2h_grant_controller.cpp
zaoxing: Fixed in fc5a0e2. `observe_progress()` now accepts the timing estimate only when `progress.counter == state.timing->absolute_close`. A counter already past it means later boundaries elapsed unseen while the drain was elsewhere, and that time cannot be charged to the window.

That covers @Samfisheryu's reproduction (next poll at counter 9 with CLOSE=3). The other variant — counter sitting exactly at CLOSE but the poll 200 ms late — the counter cannot detect at all; that one is caught by the `note_observation_gap()` signal described in the `max_safe_` thread, which marks in-flight observations unclean at both ends.

`test_late_close_observation_supplies_no_timing_estimate` pins both halves: 64 MiB issued at counter 1 / t=0, completing at counter 2 / t=5 ms, then a poll at counter 9 / t=205 ms now yields no estimate, where it previously produced 2,624 MiB. The existing `test_controller_timing_estimate_is_deterministic` still produces its estimate, so this tightened the check rather than disabling the mechanism.

native/csrc/ring/drain_thread.cpp
zaoxing: Fixed in fed8c76. One pause is outstanding at a time now: a second `pause_after_flush_and_wait()` queues on a new `pause_entry_cv_` until the first resumes, so two tokens can never coexist and the race is gone by construction. A token that is not the outstanding one — stale, or already resumed — throws `std::logic_error` instead of returning, so misuse fails where the mistake is rather than as an unrelated hang later. `record_drain_failure()` also wakes entry waiters, so a dead drain reports itself instead of blocking them.

Since `resume()` can now throw, the version-reuse path's cleanup `resume()` swallows a rejection: it runs while unwinding and must not replace the failure the caller needs to see.

`test_overlapping_pause_requests_serialise` covers the queueing, the stale-token rejection and the double-resume rejection. To be precise about what it does *not* cover: the two-token race itself needs both requests to land before the loop wakes, which I did not find a way to force deterministically. The entry gate makes that state unreachable, and the test pins the loud-failure behaviour around it.

native/csrc/ring/drain_thread.cpp
zaoxing: Fixed in fc5a0e2, taking your first suggestion: `full_count = pending_entries_` and `full_bytes = pending_bytes_` when `pending_bytes_ <= staging_free`, with the prefix walk only in the other case.

The fast path is equivalent because `pending_bytes_` is exactly the aligned sum of `scanned_` (`scan_ready` adds `align_up(actual_bytes)`, `trim_scanned` subtracts the flushed aligned sum) and `pending_entries_ == scanned_.size()`. The `flush_bytes <= staging_free` invariant that the un-waited `enqueue_d2h` in this path relies on still holds in both branches — staging free bytes only grow between the decision and the copy, since only the drain thread consumes them.

native/csrc/ring/d2h_window_grant_policy.cpp
zaoxing: Not fixed — I left this one to you deliberately, since it turns on a decision a reviewer should not make unilaterally.

The capping half changes public surface and tuning defaults: either a new `RecurringD2HWindowConfig` seed field (bindings, validation, and both docs) or a `staging_free / k` with an invented *k*. Choosing *k* wrong makes the feature worse rather than better, and that call is yours.

On the second half — seeding the initial `B` from the overrun's timing — I do not think it is sound, and would rather say so than implement it. For an overrun the window duration is unknown; all that is known is `window_ns < transfer_ns`, so `bytes × window_ns / transfer_ns` has no measured numerator. And `B` wants a value known unsafe, which `attempt.bytes` already is, more tightly than any estimate derived from it. Happy to be corrected if you meant something different.

One correction to the cost estimate, which may affect how you weigh this: the search does not start from `UINT64_MAX` in practice. The drain clamps the first transfer to what fits staging, so `min_unsafe_` becomes the staged bytes and the halving count is log2(staging capacity / true capacity), not log2(backlog / capacity). Still worth bounding, but the 4 GiB → 4 MiB ≈ 10 occurrences figure assumes a 4 GiB staging ring.

native/csrc/ring/recurring_d2h_grant_controller.cpp
zaoxing: Fixed in fc5a0e2. Moved to `D2HWindowDebugLogger::log_version_activation()` and called only when `debug_logger_` is set, which the subsystem constructs only under `debug_enabled`. The terminal-fallback message is left as a warning, per your note.

Visible in the test output: the native suite printed `[d2h_window] active version=` 17 times per run before and prints none now.

native/csrc/ring/d2h_window_subsystem.cpp
zaoxing: Fixed in bb36a5b. New `native/csrc/ring/cuda_check.h` holds one inline `check_cuda` plus a `check_cuda_launch` for the `cudaGetLastError()` shape; `d2h_window_progress.cpp`, `d2h_window_subsystem.cpp` and `d2h_window_marker.cu` all use it, and the header is listed in the Makefile's `HEADERS` so edits trigger a rebuild.

I left `drain_thread.cpp`'s `throw_cuda_failure` alone: it predates this PR and prefixes `DrainThread: `, so folding it in would change existing error message text outside this change's scope. Say the word if you would rather all four were unified and I will do that separately.

src/dmi/transport/ring.py
zaoxing: Fixed in 7b43d55 — the registry is keyed by `int(_native.D2HWindowProgressKind.PACKED_VERSION_COUNTER)`. The import sits inside the branch so a module-level attribute access does not force the extension load at import time.

No new test, deliberately: `tests/test_d2h_window_transport.py` is `pytest.mark.cpu` and builds its transport with `__new__`, bypassing the `__init__` where the registry lives. Covering it there would need `torch.ops.ring` — a built extension — and would regress that file's CPU-only property. I verified the key resolves to 0 against the built extension instead, so the change is behaviour-identical today and only removes the reordering hazard.

docs/config.md
zaoxing: Fixed in 7b43d55. Added the `capacity_flush_count_reset_interval_periods` row, so `docs/config.md` and `docs/integration-api-v1.md` now describe the same eight fields.

Also amended the `debug_enabled` description in the same table to mention pattern-version activation, since that line is now logged through the debug logger rather than unconditionally.

third_party/DMI-Megatron-Integration
zaoxing: Not fixed here, and I do not think it can be from this repository: the change you are asking for is in `DMI-Megatron-Integration`'s `schedule_runtime.py`, a separate repo. This PR only moves the submodule pointer.

Agreed on the diagnosis and on the shape of the fix — explicitly suspending and resuming the native window policy around forward-only phases, rather than only stopping boundary markers. The capacity flushes a validation phase produces genuinely charge against the training window's policy, and three of them are terminal.

One thing that *is* fixable here: the PR description says it pins DMI-Megatron-Integration `f416d73`, but the branch records `b9b2c49` — the revision you actually reviewed. The description needs updating.

PR #137, opened 2026-09-22
Review the Python oracle: three capture-path fixes, and tests for what was unpinned

zaoxing:
## What

A review pass over `src/dmi/` — the Python oracle. Three behaviour fixes and
six tests that pin behaviour which was previously unpinned or pinned by a stub.

The scope was chosen deliberately. `src/dmi/` is the reference implementation
every native C++ fix is verified against: it has been the measuring stick for
roughly twenty findings and had never itself been reviewed. A defect here is
worse than one in the port, because the port's conformance tests would
faithfully reproduce it.

Production diff is **52 lines across 3 files**; the remaining ~921 lines are
tests.

Based directly on `main` (#133), so it is independent of #132, #134 and #135.
No production file overlaps any of them — see Overlap below.

### The footer cache returned descriptors without re-checking the tenant bind

`reader.py` caches pack footers by content identity. On a hit it returned the
cached descriptor directly, skipping `_reject_a_foreign_tenant` — so a second
caller, under a different tenant, could read a descriptor the bind would have
refused.

The fix re-runs the location bind **on the hit** rather than widening the cache
key. Adding `object_key` to the key was considered and rejected: that leaves a
content-identity cache silently carrying a location-derived result, which is
the same defect one layer down. `_reject_a_foreign_tenant` is now
`reject_a_foreign_tenant(ref, tenants)` so the reader can call it against a
cached descriptor.

### The `date=` object-key segment was computed by float division

`pipeline.py:296` divided `captured_at_ns` by `1e9`, which **rounds**. The C++
port truncates. Now `// 1_000_000_000`.

The divergence window is **596 ns** per day boundary, measured rather than
estimated: `…999999403` agrees, `…999999404` diverges. A capture landing in
that window is written under tomorrow's `date=` partition by Python and
today's by the port — the two implementations disagree about where the object
lives, which is exactly the class of drift the conformance suite exists to
catch and could not see here.

### A manual flush could orphan its barrier at the failure seam

If the worker failed between a flush installing its barrier and the worker
reaching it, the barrier was never woken and the flush waited forever. Now the
failure handler wakes it.

## The six pins

These change no production behaviour. Each was verified red-then-green by
reverting the source under test and confirming exactly one test fails.

- Three CPU-direct record transforms had **zero** coverage.
- The dtype/shape bind was validated by a stub carrying **its own copy** of the
  check and its own message — `grep` found that message only in the stub and
  the `match=` asserting it, so the test passed with the product's guard
  deleted. The fix was to make the stub delegate.
- Flush-after-close: deleting either guard turns a prompt `False` into an
  unbounded hang.
- Flush against a failed pipeline: the reuse path could return `True` on a
  failure.
- The failure seam's **other** arm — the barrier wake — which the fix above
  left unpinned, because its new test parks flush before the barrier exists.
- The record-runtime rollback's stop-before-drop ordering.

## Checks

| Signal | Base `214bdb2` | Tip `9aaa871` |
|---|---|---|
| `make -C native cpu-goals` (forced `-B`) | 0 errors, 0 warnings | **0** |
| `pytest -m cpu` | 1462 passed, 0 skipped | **1499 passed, 0 skipped** |
| CI cpu skip gate | — | 1499 collected, 0 unexplained |
| `pytest -m "clickhouse and manual and not garage"` | 198 passed, 1 failed | **198 passed, 1 failed** |
| CI live skip gate | — | 199 collected, 0 skipped |
| `compileall src/dmi` | clean | clean |

The one live failure is pre-existing and environmental:
`test_a_role_that_cannot_see_one_object_is_told_to_grant_it_not_to_rebuild`
needs `CREATE USER`, and the local standalone ClickHouse has no access
management. It passes in CI.

The build check forces `-B` on purpose — without it, cached objects suppress
re-emission and a warning count is vacuous.

**Threading stability: 30 runs, zero flakes.** Twenty full-file runs plus ten
under 48-way CPU saturation on 32 cores. Latency rose ~1.7x with no
behavioural change, which is the actual proof: the interleavings are forced by
synchronisation primitives rather than sleeps, so they hold under load.
Per-test JUnit parsing confirmed all five interleaving tests ran in all twenty
runs and none silently deselected.

`make test-package` is **NOT MEASURABLE** on this host, and is not counted as
green: it dies inside the stdlib (`venv.EnvBuilder(with_pip=True)`, `ensurepip`
rc=127), reproduced with zero repo files involved. The parts that could
regress from source changes do pass — the wheel builds, and `_validate_archive`
returns clean. Only the venv smoke-install is unmeasured.

## Overlap

No production file here is touched by #132, #134 or #135. Three test files
collide and will need a trivial merge whichever lands second:

- `tests/test_engine_runtime_api.py` — all four branches
- `tests/test_native_pack_sink.py` — #134, #135
- `tests/test_capture_record_adapter.py`, `tests/test_record_cpu_direct.py` — #132

## Known, not fixed

- **`record_adapter.py:240-241` is dead code**, shadowed unconditionally by the
  `is_running` check at 231. Found during verification. It is a code change
  rather than a missing test, so it is out of scope for a behaviour-preserving
  pass and wants its own PR.
- **The `-Wall -Wextra` authority has a hole.** Two of the nine cpu goals
  compile without it — `bench_builder` and `_dmi_native_sink` — so "0 warnings"
  says nothing about `bench_builder.cpp`, `native_pack_sink.cpp` or
  `bindings_sink.cpp`. Pre-existing and untouched by this Python-only diff.
- **23 findings confirmed but classed `optional`**, each with its evidence.
  They are a maintainer's call, not the pass's.

## How to read the ledger

56 findings reported, 26 verified, 9 fixed: 5 refuted outright, 13
confirmed-then-downgraded, 1 deduped. The gap is the point — every refutation
and downgrade would otherwise have been a fix applied on a reviewer's say-so.
Reachability was the sole reason for the downgrades: the mechanism was real and
demonstrated, and no production caller could reach it.


Samfisheryu (approved):
Reviewed at 9aaa871. No blockers found. Independently passed 1,499 CPU tests and 53 live ClickHouse E2E/native-reader parity tests; restoring each of the three original implementations reproduced the corresponding regression. Sink-routing checks also pass. This PR leaves the shared Ring, sink-selection boundary, and existing production ClickHouse path unchanged. The Python reader fixes also apply when reading native-written capture packs. GPU E2E was not rerun for this Python-only production diff.

